PR #5, opened 2019-01-09
Feature/public profile (Left sidebar user information)

leomendoza123:


leomendoza123:
[Angular prototype: Display of “left” side bar user general information.](https://trello.com/c/No5yUy3d/5563-angular-prototype-display-of-left-side-bar-user-general-information)

PR #24, opened 2019-07-08
Hide second level button on home page

leomendoza123:


leomendoza123:
[Hide second level menu on home page](https://trello.com/c/UO73pZcZ/6033-hide-second-level-menu-on-home-page)

PR #26, opened 2019-07-10
Support safari flexbox

leomendoza123:


leomendoza123:
[iPad devices older than iPad Air2 do not show the menu.](https://trello.com/c/QiwqjZED/6074-ipad-devices-older-than-ipad-air2-do-not-show-the-menu)

PR #36, opened 2019-08-01
Fix/search bar focus state

leomendoza123:


leomendoza123:
[Search bar states](https://trello.com/c/f9hwF5jE/6105-search-bar-states)

PR #40, opened 2019-09-24
Fix/background palette a11y

leomendoza123:
Please first merge #38


leomendoza123:
[New front page: colour contrast on registry search drop down insufficient.](https://trello.com/c/cQmp9HD3/6287-new-front-page-colour-contrast-on-registry-search-drop-down-insufficient)

PR #41, opened 2019-09-25
Feature/story book

leomendoza123:
Please merge #38 first

PR #43, opened 2019-10-15
Feature/init bit

leomendoza123:
please merge #41 first


leomendoza123:
[Replace works "Load More" with Pagination (5)](https://trello.com/c/Z9l0BLEp/6237-replace-works-load-more-with-pagination-5)

PR #49, opened 2019-11-11
Feature/sentence case hero title

leomendoza123:


leomendoza123:
[New Homepage: switch text to normal case](https://trello.com/c/216HpPee/6389-new-homepage-switch-text-to-normal-case)

PR #55, opened 2019-11-13
Add active state to the nav menu

leomendoza123:
- add grey color to the palette


leomendoza123:
[On the home page the current location is not highlighted on the main menu](https://trello.com/c/z56jThzf/6423-on-the-home-page-the-current-location-is-not-highlighted-on-the-main-menu)

PR #56, opened 2019-11-13
Add dir RTL HTML flag dynamically

leomendoza123:


leomendoza123:
[Set RTL HTML flag dynamically](https://trello.com/c/fA2Ftcvx/6422-set-rtl-html-flag-dynamically)

PR #63, opened 2019-11-21
Fix sign in mobile button

leomendoza123:


leomendoza123:
[Sign in register button on Mobile menu in ios and safari does not work](https://trello.com/c/pteE0LBb/6424-sign-in-register-button-on-mobile-menu-in-ios-and-safari-does-not-work)

PR #64, opened 2019-11-21
Feature/unique index file length

leomendoza123:


leomendoza123:
[New Homepage: Language switcher does not work correctly](https://trello.com/c/mqCzpgOK/6416-new-homepage-language-switcher-does-not-work-correctly)

leomendoza123:
[NGINX mapping keeps pointing to the `en/index.html` even when the language cookie was changed](https://trello.com/c/jide0pv2/6412-nginx-mapping-keeps-pointing-to-the-en-indexhtml-even-when-the-language-cookie-was-changed)

PR #68, opened 2019-11-26
Use woff2 instead of woff

leomendoza123:


leomendoza123:
[Prioritize woff2](https://trello.com/c/aoWO8eUv/6439-prioritize-woff2)

PR #69, opened 2019-11-26
Fix focus register url on tab

leomendoza123:


leomendoza123:
[New homepage: VO does not announce the register now link under "1 Register " heading properly in safari](https://trello.com/c/NKzCUYcy/6432-new-homepage-vo-does-not-announce-the-register-now-link-under-1-register-heading-properly-in-safari)

PR #72, opened 2019-12-05
Fix search box gap

leomendoza123:


leomendoza123:
[New homepage: Safari has a gap between search and registry button.](https://trello.com/c/tSdBo7oG/6445-new-homepage-safari-has-a-gap-between-search-and-registry-button)

PR #74, opened 2019-12-09
Change language selector standard to underscore

leomendoza123:


leomendoza123:
[New homepage: Chinese Langs (both) do not work.](https://trello.com/c/zvoRzrgh/6452-new-homepage-chinese-langs-both-do-not-work)

PR #75, opened 2019-12-10
Add unique length post build script docs

leomendoza123:
- Update language files


src/locale/messages.xlf
```diff
@@ -1297,14 +1298,6 @@
         <source>Account Settings</source>
       </segment>
     </unit>
-    <unit id="workspace.developer_tools">
```
amontenegro: Is renoving this ok?

leomendoza123: Yes since is not been removed, just moved into line 152.

leomendoza123:
[new home page: language menu doesn't work with Cloudfare](https://trello.com/c/DFPmoYme/6451-new-home-page-language-menu-doesnt-work-with-cloudfare)

PR #82, opened 2019-12-19
Fix broken help urls

leomendoza123:


leomendoza123:
[New home page: Incorrect links for nav menu items that point to external sites](https://trello.com/c/out2geN7/6480-new-home-page-incorrect-links-for-nav-menu-items-that-point-to-external-sites)

PR #86, opened 2020-01-08
Enable the site to work with the backend through nginx

amontenegro:


amontenegro:
@lizkrznarich @leomendoza123 please review this along with https://github.com/ORCID/registry_vagrant/pull/500

Thanks

PR #89, opened 2020-01-15
Remove invalid regexp and relies on Angular validator and the require…

leomendoza123:
…ment of a TLD for emails


leomendoza123:
[Reset password/what is my iD thinks email addresses with plus signs in them are invalid](https://trello.com/c/BeaoUBR4/6502-reset-password-what-is-my-id-things-email-addresses-with-plus-signs-in-them-are-invalid)

PR #113, opened 2020-02-20
Add comma separeted results

leomendoza123:


leomendoza123:
[Search results don't have space between each value](https://trello.com/c/1TKt0K7q/6571-search-results-dont-have-space-between-each-value)

PR #115, opened 2020-02-21
Fix pub api endpoint

amontenegro:


amontenegro:
@leomendoza123 please review, I think we should this for next prod release.

Thanks

PR #117, opened 2020-02-24
Fix search a11y

leomendoza123:


leomendoza123:
[New app, search accessibility](https://trello.com/c/UNo1bNhT/6569-new-app-search-accessibility)

PR #120, opened 2020-02-27
Feature/analytics internal navigation

leomendoza123:


leomendoza123:
[New app, page loading speed on analytics](https://trello.com/c/OPphupDN/6565-new-app-page-loading-speed-on-analytics)

PR #121, opened 2020-02-28
Fix typo, however, we should also fix the component name

amontenegro:


amontenegro:
[Typo "advance search"](https://trello.com/c/qR4qzvRg/6587-typo-advance-search)

PR #122, opened 2020-02-28
Extract id when searching by orcid

lizkrznarich:


lizkrznarich:
[Advanced search with full URL in ORCID iD field](https://trello.com/c/vpHQBWMz/6589-advanced-search-with-full-url-in-orcid-id-field)

PR #130, opened 2020-03-12
Feature/clone orcid source translations

leomendoza123:


leomendoza123:
[Create a script to automatically move translations values from the old app to the new app](https://trello.com/c/7pktwCnH/6597-create-a-script-to-automatically-move-translations-values-from-the-old-app-to-the-new-app)

PR #138, opened 2020-03-20
Feature/clone values with most translations

leomendoza123:


leomendoza123:
[Update the orcid angular prebuild script](https://trello.com/c/DJxijTlc/6598-update-the-orcid-angular-prebuild-script)

PR #141, opened 2020-03-25
Update ngx-cookie-service to fix the issue

leomendoza123:


leomendoza123:
[Orcid Angular banners cookies are SameSite=None and not secure](https://trello.com/c/KjEiCXvR/6634-orcid-angular-banners-cookies-are-samesitenone-and-not-secure)

PR #152, opened 2020-04-19
Feature/angular 9

leomendoza123:
Update to Angular 9 

- lightweight builds (reduce the build sizes) 
- faster on run time with the new Ivy Angular engine
- make releases faster by building all the languages in on run (instead of on  12) 
- improve error reports for developers 
- add static translations for .ts files
- remove complicated build scripts on package.json


leomendoza123:
[Update to Angular 9](https://trello.com/c/RyM1ihMA/6686-update-to-angular-9)

PR #155, opened 2020-04-23
Feature/register

leomendoza123:


leomendoza123:
[Register (13)](https://trello.com/c/SK0r2MHd/6544-register-13)

src/app/register/register.module.ts
```diff
@@ -0,0 +1,59 @@
...
+import { StepAComponent } from './components/step-a/step-a.component'
+import { StepBComponent } from './components/step-b/step-b.component'
+import { StepCComponent } from './components/step-c/step-c.component'
+import { FormPersonalComponent } from './components/form-personal/form-personal.component'
+import { FormPasswordComponent } from './components/form-password/form-password.component'
+import { FormNotificationsComponent } from './components/form-notifications/form-notifications.component'
+import { FormVisibilityComponent } from './components/form-visibility/form-visibility.component'
+import { FormTermsComponent } from './components/form-terms/form-terms.component'
+import { FormsModule, ReactiveFormsModule } from '@angular/forms'
+// tslint:disable-next-line: max-line-length
```
DanielPalafox, line 14: @leomendoza123 I think we should avoid disabling tslint rules and put the code in different lines to avoid error rule max-line-length

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  addAdditionalEmail(): void {
+    const controlName = (
+      Object.keys(this.additionalEmails.controls).length + 1
+    ).toString()
+    this.additionalEmails.addControl(
+      controlName,
+      new FormControl('', {
+        validators: [Validators.email, Validators.pattern(TLD_REGEXP)],
+        // asyncValidators: this._register.backendValueValidate(controlName),
+      })
```
DanielPalafox, line 56: @leomendoza123 erase if it is not used

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.html
```diff
@@ -0,0 +1,102 @@
...
+  class="mat-button-font"
+>
+  <mat-icon>add</mat-icon>
+  <ng-container i18n="@@register.additionalEmailAdd"
+    >Add another email</ng-container
+  >
+</a>
+
+<mde-popover #namesPopover="mdePopover" [mdePopoverOverlapTrigger]="false">
+  <mat-card style="max-width: 300px" class="mat-elevation-z3">
```
DanielPalafox, line 85: @leomendoza123 I would put all the styles in the css files

src/app/register/register.module.ts
leomendoza123: Good observation but also ts lint says 

```
Line breaks are not allowed in import declaration (import-spacing)
```

So I need to ignore a rule anyway

src/app/register/components/backend-error/backend-error.component.html
```diff
@@ -0,0 +1,24 @@
...
+</ng-container>
+
+<ng-container
+  *ngIf="recognizedError[errorCode] === 2"
+  i18n="@@register.additionalEmailMatch"
+>
+  Additional emails can not be duplicated
+</ng-container>
+
+<ng-container *ngIf="unrecognizedError"> {{ errorCode }} </ng-container>
```
DanielPalafox, line 24: @leomendoza123 I think we were using the icon error to denote are errors from the server

src/app/register/components/backend-error/backend-error.component.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, OnInit, Input } from '@angular/core'
+
+enum RegisterBackendErrors {
+  'orcid.frontend.verify.duplicate_email',
+  'additionalEmailCantBePrimaryEmail',
+  'duplicatedAdditionalEmail',
```
DanielPalafox, line 6: @leomendoza123 the property 'orcid.frontend.verify.deactivated_email' is missing and when you try to register with an unclaimed error throws the next message 'unclaimed@test.com already exists in our system as an unclaimed record. Would you like to <a href='https://dev.orcid.org/resend-claim?email=unclaimed%40test.com'>resend the claim email</a>?'

src/app/register/components/form-password/form-password.component.html
```diff
@@ -0,0 +1,176 @@
...
+</ng-template>
+
+<ng-template #invalid>
+  <mat-icon class="material-icons-outlined">check_circle</mat-icon>
+  <!--
+    <img
+      src="assets/vectors/mat-baseline-check_circle_outline.svg"
+      alt="orcid logo"
+    />
+  -->
```
DanielPalafox, line 127: @leomendoza123 Remove this unused lines

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.html
```diff
@@ -0,0 +1,101 @@
...
+        *ngIf="
+          additionalEmails.hasError('email', additionalEmailControl.key) ||
+          additionalEmails.hasError('pattern', additionalEmailControl.key)
+        "
+        i18n="@@register.invalidEmail"
+      >
+        Invalid email format
+      </mat-error>
+      <!--
+        <mat-error
```
DanielPalafox, line 38: @leomendoza123 remove unused lines

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.html
```diff
@@ -0,0 +1,101 @@
...
+      [mdePopoverTriggerFor]="namesPopover"
+      mdePopoverTriggerOn="click"
+      #namesPopoverTrigger="mdePopoverTrigger"
+    >
+      <mat-icon>info</mat-icon>
+    </button>
+  </div>
+</ng-container>
+
+<a (click)="addAdditionalEmail()" class="mat-button-font">
```
DanielPalafox, line 76: @leomendoza123 I think there should as well an option to remove the additional emails in case the user changes his mind

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.html
```diff
@@ -0,0 +1,101 @@
...
+    <ng-container i18n="@@register.additionalEmailInfo"
+      >Ensure that you never lose access to your ORCID record by registering all
+      of your email addresses on your account.</ng-container
+    >
+
+    <button
+      mat-button
+      color="primary"
+      aria-label="close"
+      (click)="namesPopoverTrigger.closePopover()"
```
DanielPalafox, line 94: @leomendoza123 throws an error in the console when clicked it

src/app/register/components/form-personal/form-personal.component.html
```diff
@@ -0,0 +1,160 @@
...
+        >
+          <app-backend-error [errorCode]="error"></app-backend-error>
+        </div>
+      </mat-error>
+    </mat-form-field>
+    <mat-form-field appearance="outline">
+      <mat-label i18n="@@register.confirmEmail"
+        >Confirm primary email</mat-label
+      >
+      <input formControlName="confirmEmail" matInput (paste)="(false)" />
```
DanielPalafox, line 89: @leomendoza123 I will remove the 'paste' attribute since we already have the validation if they are not the same email's and some users have very long email's

src/app/register/components/form-personal/form-personal.component.ts
```diff
@@ -0,0 +1,166 @@
...
+      validators: [Validators.email, Validators.pattern(TLD_REGEXP)],
+    }),
+  })
+
+  ngOnInit() {
+    this.emails = new FormGroup(
+      {
+        email: new FormControl('', {
+          validators: [
+            Validators.required,
```
DanielPalafox, line 64: @leomendoza123 I think it would be a good idea to create a validator for emails which include Validators.required, email and TLD_REGEXP to reuse it whenever is needed

src/app/register/components/form-terms/form-terms.component.html
```diff
@@ -0,0 +1,34 @@
+<h3 class="mat-subheading-2" i18n="@@register.termsOfUse">Terms of Use</h3>
+
+<ng-container [formGroup]="form">
+  <mat-checkbox
+    class="orcid-checkbox-baseline"
+    color="primary"
+    formControlName="termsOfUse"
+    ><ng-container i18n="@@register.consentTo">I consent to the </ng-container
```
DanielPalafox, line 8: @leomendoza123 the attribute preservedWitheSpaces won't work with ng-container you should use p instead

src/app/register/components/form-visibility/form-visibility.component.html
```diff
@@ -0,0 +1,48 @@
...
+  <p i18n="@@register.visibilityDescription3">
+    By default, who should be able to see information added to your ORCID
+    Record?
+  </p>
+  <p>
+    <mat-radio-group
+      formControlName="activitiesVisibilityDefault"
+      color="primary"
+    >
+      <mat-radio-button
```
DanielPalafox, line 28: @leomendoza123 Could you add a margin between the options they are to close to each other and the icons are missing

src/app/register/components/form-visibility/form-visibility.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      provide: NG_ASYNC_VALIDATORS,
+      useExisting: forwardRef(() => FormVisibilityComponent),
+      multi: true,
+    },
+  ],
+})
+export class FormVisibilityComponent extends BaseForm implements OnInit {
+  visibilityOptions = ['PUBLIC', 'LIMITED', 'PRIVATE']
```
DanielPalafox, line 32: @leomendoza123 I think we need to add these on the constants files since we will reuse it later

src/app/register/pages/register/register.component.ts
```diff
@@ -0,0 +1,143 @@
...
+      this._register
+        .checkDuplicatedResearcher({
+          familyNames: this.personalData.familyNames.value,
+          givenNames: this.personalData.givenNames.value,
+        })
+        .subscribe((value) => {
+          if (value.length > 0) {
+            this.openDialog(value)
+          } else {
+            // continueRegistration
```
DanielPalafox, line 96: @leomendoza123 this else statement is useless

src/app/register/pages/register/register.component.html
```diff
@@ -0,0 +1,78 @@
...
+          <app-step-a [(formGroup)]="FormGroupStepA"></app-step-a>
+        </mat-step>
+        <mat-step [stepControl]="FormGroupStepB">
+          <ng-template matStepLabel i18n="@@register.personalSecurityVisibility"
+            >Security and visibility</ng-template
+          >
+          <app-step-b
+            [(formGroup)]="FormGroupStepB"
+            [personalData]="personalData"
+            (submit)="submitStepB()"
```
DanielPalafox, line 64: @leomendoza123 I think you should remove this since inside the component there is a button with matStepperNext

src/app/register/pages/register/register.component.html
```diff
@@ -64,12 +65,13 @@
           ></app-step-b>
         </mat-step>
         <mat-step [stepControl]="FormGroupStepC" #lastStep>
-          <ng-template matStepLabel i18n="@@register.personalNotificationsTerm"
-            >Notifications and terms</ng-template
+          <ng-template matStepLabel i18n="@@register.VisibilityAndTerm"
```
DanielPalafox, line 68: @leomendoza123 I think the property it should be 'register.visibilityAndTerm' and register.securityAndNotifications using camelCase

src/app/register/components/backend-error/backend-error.component.ts
leomendoza123: This is expected behavior, I am adding the following comment to explain better on the code, please let me know if this is clear enough 

```
    // When the error text is not listed on the RegisterBackendErrors enum
    // the error message will be displayed as it comes from the backend
    // This is because the backend might return code or a text ready for the UI 
```

DanielPalafox:
@leomendoza123 I checked the code and it looks very good. Great work!. I left you some comments

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.html
leomendoza123: That makes sense, with this PR I am just cloning current functionality, where they can just leave the input empty if they do not want to add additional emails. 

Let's ask Dam about this at our next meeting.

leomendoza123: oh wait, the old app also have this delete button, I will add it 👍

src/app/register/components/form-personal/form-personal.component.html
leomendoza123: This is a feature required by Paula, she doesn't want to allow users to paster the email on the confirmation field, this is to make sure the user writes the email there.

src/app/register/components/form-personal/form-personal.component.ts
leomendoza123: That might help sometimes, but also: 

a- keeping the validator on separate instances can help to improve readability on what's been validated for each input 
b- in the future form we would need to use the email validation without it been required 

What do you think?

src/app/register/components/form-visibility/form-visibility.component.ts
leomendoza123: true!

src/app/register/pages/register/register.component.html
leomendoza123: 👍

src/app/register/pages/register/register.component.html
leomendoza123: 👍

src/app/register/components/backend-error/backend-error.component.html
leomendoza123: That's not really a pattern we follow but I  have just open an issue on Orcid Angular to address this with the product team

src/app/register/components/form-visibility/form-visibility.component.html
leomendoza123: Perfect added the margins, and we have a card for the icons https://trello.com/c/Q7sMADQg/6744-register-add-visibility-icons

src/app/register/components/form-terms/form-terms.component.html
leomendoza123: good catch, the problem was that there was no space so I can still use ng-container so I don't add any extra HTML tag on the DOM.

leomendoza123: I have fixed this by adding spaces

src/app/register/components/form-visibility/form-visibility.component.html
leomendoza123: I have added the margin, and there is a card for the icons since those need to be defined by Dan https://trello.com/c/Q7sMADQg/6744-register-add-visibility-icons

src/app/register/components/form-personal-additional-emails/form-personal-additional-emails.component.html
leomendoza123: I have created this issue to solve this one https://github.com/ORCID/orcid-angular/issues/182

leomendoza123: This doesn't produce any usability limitation

src/app/register/components/backend-error/backend-error.component.ts
DanielPalafox: Ok but I think we need to resize the area of the error messages if it's too long as the example I put and the property 'orcid.frontend.verify.deactivated_email' is a known error and it should be in the enum

leomendoza123:
[Register - Add google captcha support](https://trello.com/c/0kOyQpsH/6711-register-add-google-captcha-support)

PR #162, opened 2020-04-27
6699 qa signin unable to use https or http full url as username

DanielPalafox:
https://trello.com/c/Kf9ayPZ3/6699-qa-signin-unable-to-use-https-or-http-full-url-as-username

### DIFF
--- a/src/app/constants.ts
+++ b/src/app/constants.ts
@@ -47,4 +47,5 @@ export const GRID_COLUMNS = {
 // this REGEXP adds the requirement of ending with a TLD as defined on RFC2396
 export const TLD_REGEXP = /^.*\.([a-zA-Z\-])([a-zA-Z\-]{0,61})([a-zA-Z\-])$/
 export const ORCID_REGEXP = /(\d{4}-){3,}\d{3}[\dX]$/i
-export const ORCID_URI_REGEXP = /(http[s]?:\/\/)(orcid\.org\/|qa\.orcid\.org\/|dev\.orcid\.org\/|localhost.*)(\d{4}-){3,}\d{3}[\dX]$/
+// https://regex101.com/r/V95col/1
+export const ORCID_URI_REGEXP = /(http[s]?:\/\/)(orcid\.org\/|qa\.orcid\.org\/|sandbox\.orcid\.org\/|dev\.orcid\.org\/|localhost.*)(\d{4}-){3,}\d{3}[\dX]$/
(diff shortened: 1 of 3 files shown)


DanielPalafox:
[[QA] Signin - unable to use https or http full URL as username](https://trello.com/c/Kf9ayPZ3/6699-qa-signin-unable-to-use-https-or-http-full-url-as-username)

DanielPalafox:
[[QA] Signin - unable to use https or http full URL as username](https://trello.com/c/Kf9ayPZ3/6699-qa-signin-unable-to-use-https-or-http-full-url-as-username)

src/app/constants.ts
```diff
@@ -43,3 +47,4 @@ export const GRID_COLUMNS = {
 // this REGEXP adds the requirement of ending with a TLD as defined on RFC2396
 export const TLD_REGEXP = /^.*\.([a-zA-Z\-])([a-zA-Z\-]{0,61})([a-zA-Z\-])$/
 export const ORCID_REGEXP = /(\d{4}-){3,}\d{3}[\dX]$/i
+export const ORCID_URI_REGEXP = /(http[s]?:\/\/)(orcid\.org\/|qa\.orcid\.org\/|dev\.orcid\.org\/|localhost.*)(\d{4}-){3,}\d{3}[\dX]$/
```
leomendoza123, line 50: Perfect! 

I am going to create another issue to implement this on the search as well, and let's keep reusing/improving this reg for URLs. Just two changes here: 

a- please add support to sandbox
b- I am adding a testing URL as a comment for each regexp on the `feature/register` PR, look:
https://github.com/ORCID/orcid-angular/blob/938b63671f7aec35bbf657353193fd97ab606104/src/app/constants.ts#L42-L55
So please add a URL like this one https://regex101.com/r/6U0wA9/1/ (with sandbox) as a comment on top of you regexp

DanielPalafox: @leomendoza123 Ok. Sounds great!

src/app/constants.ts
```diff
@@ -16,6 +16,10 @@ export function isValidOrcidFormat(id) {
   return id && regExp.test(id)
 }
 
+export function getOrcidNumber(orcid) {
+  return orcid.match(ORCID_REGEXP)[0]
```
amontenegro, line 20: @DanielPalafox should this be using the `ORCID_URI_REGEXP`? instead of the ORCID_REGEXP

DanielPalafox: @amontenegro No, necessarily. You can use both expressions to achieve this but I am using ORCID_REGEXP because it will look specifically for the Orcid id and in the other part the ORCID_URI_REGEXP it is to validate the domain and the Orcid id

PR #164, opened 2020-04-27
Encode special characters in sigin

DanielPalafox:
https://trello.com/c/0GP6CtMV/6701-qa-signin-unable-to-sign-in-with-email-addresses-containing-a-plus-sign


DanielPalafox:
[[QA] Signin - unable to sign in with email addresses containing a plus sign](https://trello.com/c/0GP6CtMV/6701-qa-signin-unable-to-sign-in-with-email-addresses-containing-a-plus-sign)

src/app/core/custom-encoder/custom-encoder.component.ts
```diff
@@ -0,0 +1,16 @@
+import { HttpParameterCodec } from '@angular/common/http'
+
+export class CustomEncoderComponent implements HttpParameterCodec {
+  encodeKey(key: string): string {
```
leomendoza123, line 4: This is a great solution! Do you think it makes sense to change the postfix from `.component.ts ` to `.encoder.ts`? since this is not a component or is not a service, but just a class

leomendoza123: Also, remove the testing... I believe it would break because it won't be able to `compileComponents` or `createComponent`  a component.

DanielPalafox: @leomendoza123 Yes, it's sound better. encoder.ts. Ok, I will remove the test.

PR #170, opened 2020-04-30
Fix/chinese language code

leomendoza123:


leomendoza123:
[Can't change to Chinise languages in the new app](https://trello.com/c/mpd7dxcQ/6719-cant-change-to-chinise-languages-in-the-new-app)

PR #172, opened 2020-05-04
Fix Chinese language folder names

leomendoza123:


leomendoza123:
[Can't change to Chinese languages in the new app](https://trello.com/c/mpd7dxcQ/6719-cant-change-to-chinese-languages-in-the-new-app)

package.json
```diff
@@ -25,13 +25,13 @@
...
     "prebuild": "rimraf dist && yarn build:i18n && yarn build:browserslist",
-    "build": "yarn build:prod",
+    "build": "ng build --prod --localize",
     "postbuild": "ts-node -P scripts/tsconfig.json scripts/postbuild.ts --env production",
     "build:i18n:clone": "ts-node -P scripts/tsconfig.json scripts/translate-file-clone.prebuild.ts",
     "build:i18n:extract": "ng xi18n  --i18n-format=xlf2 --output-path locale",
     "build:i18n:generate": "ts-node -P scripts/tsconfig.json scripts/translate-file-generator.prebuild.ts",
     "build:i18n": "yarn build:i18n:extract && yarn build:i18n:generate",
-    "build:prod": "yarn prebuild && ng build --prod --localize && yarn postbuild",
+    "build:prod": "yarn build",
```
DanielPalafox, line 34: @leomendoza123 I think we should leave it as it was before to run prebuild and postbuild as well or otherwise remove them also from qa and sandbox

leomendoza123: @DanielPalafox Good observation. Npm by default runs prebuild and post-build when you run `npm run build`. So With this I am just simplifying this script

PR #175, opened 2020-05-05
Feature/add error message url

leomendoza123:


leomendoza123:
["Register" and "Recover account details" buttons are unresponsive](https://trello.com/c/3StbCA0S/6630-register-and-recover-account-details-buttons-are-unresponsive)

PR #176, opened 2020-05-05
Display length of authentication and recovery code in mat error

DanielPalafox:
https://trello.com/c/r0gAeSRZ/6725-qa-signin-2fa-character-counter-disappears-when-error-message-is-displayed

### DIFF
--- a/src/app/sign-in/components/two-factor/two-factor.component.scss
+++ b/src/app/sign-in/components/two-factor/two-factor.component.scss
@@ -6,6 +6,16 @@
   margin-top: 12px;
 }
 
+.error-message {
+  float: left;
+  margin: 0;
+}
+
+.error-length {
+  float: right;
+  margin: 0;
+}
+
 .orcid-error {
   margin-top: 20px !important;
 }
--- a/src/app/sign-in/components/two-factor/two-factor.component.scss
+++ b/src/app/sign-in/components/two-factor/two-factor.component.scss
@@ -1,3 +1,7 @@
+:host {
+  width: 100%;
+}
+
 .mat-button-font {
   margin: auto;
 }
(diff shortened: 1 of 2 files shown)


src/app/sign-in/components/two-factor/two-factor.component.scss
```diff
@@ -6,6 +6,16 @@
   margin-top: 12px;
 }
 
+.error-message {
+  float: left;
```
leomendoza123, line 10: I believe this is going to give you problems on Arabian

leomendoza123: Test it please and if that's the case please create an overwrite for RTL like this 

https://github.com/ORCID/orcid-angular/blob/5b629161eab8f0427135bc0641f3f4553d0e86e3/src/app/home/pages/home/home.component.scss#L90-L92

This will apply for whenever you use explicit `right` or `left` values

DanielPalafox:
[[QA] Signin: 2FA character counter disappears when error message is displayed](https://trello.com/c/r0gAeSRZ/6725-qa-signin-2fa-character-counter-disappears-when-error-message-is-displayed)

leomendoza123:
@DanielPalafox One more comment about this one, I really like the idea of adding the related Trello card id to the branch name. What do you think if we keep using `feature/<card-id>-<branch-name>` and `fix/<card-id>-<branch-name>`?

src/app/sign-in/components/two-factor/two-factor.component.scss
DanielPalafox: Ok

DanielPalafox:
@leomendoza123 Ok works for me.

PR #183, opened 2020-05-11
Add given and family names to advanced search

DanielPalafox:
https://trello.com/c/kjrf6tcV/6629-advance-search-doesnt-return-diacritics

### DIFF
--- a/src/app/core/search/search.service.ts
+++ b/src/app/core/search/search.service.ts
@@ -5,7 +5,8 @@ import { Observable } from 'rxjs'
 import { SearchParameters, SearchResults } from 'src/app/types'
 import { ErrorHandlerService } from '../error-handler/error-handler.service'
 import { catchError } from 'rxjs/operators'
-import { ORCID_REGEXP } from 'src/app/constants'
+import { containsDiacritics, ORCID_REGEXP } from 'src/app/constants'
+
 @Injectable({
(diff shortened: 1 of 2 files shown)


DanielPalafox:
[Advance search doesn't return diacritics](https://trello.com/c/kjrf6tcV/6629-advance-search-doesnt-return-diacritics)

src/app/core/search/search.service.ts
```diff
@@ -56,13 +56,15 @@ export class SearchService {
       const searchParameters = []
       if (escapedParams.firstName) {
         let searchValue = `given-names:${escapedParams.firstName}`
+        searchValue += ` OR given-and-family-names:${escapedParams.firstName}`
         if (escapedParams.otherFields === 'true') {
```
leomendoza123, line 60: @amontenegro @dandineen 

So if we are using `given-names:X OR given-and-family-names:` for given names, and `family-name:X given-and-family-names:X` for family names. 

Isn't it more simple to just use `given-and-family-names` for both name and family name inputs?

leomendoza123: As you have proposed on slack please use the given-and-family-names for inputs with diacritics

DanielPalafox: Done!

PR #186, opened 2020-05-12
Improve CSRF error management

leomendoza123:


leomendoza123:
[Add more logging for CSRF exceptions](https://trello.com/c/s1lmJnYo/6746-add-more-logging-for-csrf-exceptions)

src/app/core/error-handler/error-handler.service.ts
```diff
@@ -1,13 +1,24 @@
...
 })
 export class ErrorHandlerService {
-  constructor() {}
+  browserSupport = 'unchecked'
+  constructor(
+    private _platform: PlatformInfoService,
+    private _cookie: CookieService
+  ) {
+    this.checkBrowser()
+    this._cookie.set('XSRF-TOKEN', 'test')
```
DanielPalafox, line 18: I think you should put a todo tag or comment here to remove it later. In case of someone else continues with te fix

leomendoza123: oh god this was just for testing... removing

PR #187, opened 2020-05-13
Add Google Crash Analytics 

leomendoza123:
Instead of waiting for users to report about the errors they get. This feature to allow us to get reports of what toaster errors are been shown to the users, for instance: 

image

A pitfall of this is that we are already over the Google Analytics amount of allowed hits per month. And this will add more hits: approximately 0.7 multiply by the amount of toasters users get


leomendoza123:
[Crash Analytics](https://trello.com/c/CLEiFurV/6827-crash-analytics)

PR #194, opened 2020-05-14
Remove default visibility selection

leomendoza123:


leomendoza123:
[Angular Register Form: the visibility selector should not have a default value selected](https://trello.com/c/KTsv3L2I/6748-angular-register-form-the-visibility-selector-should-not-have-a-default-value-selected)

PR #208, opened 2020-05-20
Fix screen direction

leomendoza123:


leomendoza123:
[[QA] Angular pages shown as RTL](https://trello.com/c/37k23rD4/6766-qa-angular-pages-shown-as-rtl)

PR #209, opened 2020-05-20
fix ie11 sign error with svg icons

DanielPalafox:
https://trello.com/c/5LsWFAl4/6759-qa-angular-signin-is-broken-on-ie11

### DIFF
--- a/src/app/sign-in/components/social/social.component.html
+++ b/src/app/sign-in/components/social/social.component.html
@@ -4,9 +4,7 @@
   class="institution mat-button-wrap-text-with-icon mat-elevation-z0"
   type="button"
 >
-  <a>
-    <img src="./assets/vectors/institutional-access.svg" alt="Institutional" />
-  </a>
+  <img src="./assets/vectors/institutional-access.svg" alt="Institutional" />
   <ng-container i18n="@@ngOrcid.signin.institutionalAccount">
     Access through your institution
   </ng-container>
@@ -17,9 +15,7 @@
   class="google mat-button-wrap-text-with-icon mat-elevation-z0"
   type="button"
 >
-  <a>
-    <img src="./assets/vectors/google.svg" alt="Google" />
-  </a>
+  <img src="./assets/vectors/google.svg" alt="Google" />
   <ng-container i18n="@@ngOrcid.signin.google">
     Sign in with Google
   </ng-container>
@@ -30,9 +26,7 @@
   class="facebook mat-button-wrap-text-with-icon mat-elevation-z0"
   type="button"
 >
-  <a>
-    <img src="./assets/vectors/facebook.svg" alt="Facebook" />
-  </a>
+  <img src="./assets/vectors/facebook.svg" alt="Facebook" />
   <ng-container i18n="@@ngOrcid.signin.facebook">
     Sign in with Facebook
   </ng-container>
--- a/src/app/sign-in/components/social/social.component.html
+++ b/src/app/sign-in/components/social/social.component.html
@@ -4,10 +4,10 @@
   class="institution mat-button-wrap-text-with-icon mat-elevation-z0"
   type="button"
 >
-  <mat-icon svgIcon="institutional-access"></mat-icon>
-  <ng-container
-    i18n="@@ngOrcid.signin.institutionalAccount"
-  >
+  <a>
+    <img src="./assets/vectors/institutional-access.svg" alt="Institutional" />
+  </a>
+  <ng-container i18n="@@ngOrcid.signin.institutionalAccount">
     Access through your institution
   </ng-container>
 </button>
@@ -17,7 +17,9 @@
   class="google mat-button-wrap-text-with-icon mat-elevation-z0"
   type="button"
 >
-  <mat-icon svgIcon="google"></mat-icon>
+  <a>
+    <img src="./assets/vectors/google.svg" alt="Google" />
+  </a>
   <ng-container i18n="@@ngOrcid.signin.google">
     Sign in with Google
   </ng-container>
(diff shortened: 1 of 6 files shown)


DanielPalafox:
[[QA] Angular signin is broken on IE11](https://trello.com/c/5LsWFAl4/6759-qa-angular-signin-is-broken-on-ie11)

src/assets/scss/material.orcid.overwrites.scss/button.scss
```diff
@@ -21,10 +21,13 @@ button.mat-button-wrap-text-with-icon {
   border: solid 1px rgba(0, 0, 0, 0.12);
   margin-top: 12px;
 
-  mat-icon {
+  a {
     position: absolute;
     left: 5%;
```
leomendoza123, line 26: Let's not use percentage for these margins, this will make the button look unaligned on some screen sizes. 

image

Also for Arabic, we might rather have these icons on the right, some docs on material about this https://material.io/design/usability/bidirectionality.html#mirroring-elements

DanielPalafox: Done

src/assets/scss/material.orcid.overwrites.scss/button.scss
```diff
@@ -23,8 +23,36 @@ button.mat-button-wrap-text-with-icon {
...
+      white-space: normal;
+      line-height: 24px;
+      min-height: 36px;
+    }
+  }
+}
+
+[dir='rtl'] button.mat-button-wrap-text-with-icon {
+  width: 100%;
+  border: solid 1px rgba(0, 0, 0, 0.12);
```
leomendoza123, line 48: The colors should be declared on the `_button-theme.scss` file to reuse the palette colors, 
We have an open issue about this https://github.com/ORCID/orcid-angular/issues/219

src/app/sign-in/components/social/social.component.html
```diff
@@ -4,10 +4,10 @@
   class="institution mat-button-wrap-text-with-icon mat-elevation-z0"
   type="button"
 >
-  <mat-icon svgIcon="institutional-access"></mat-icon>
-  <ng-container
-    i18n="@@ngOrcid.signin.institutionalAccount"
-  >
+  <a>
```
leomendoza123, line 7: Is a little odd to use an anchor tag, to just contain an element (not linked to anything). 
But I don't think this is critical we can keep it

src/assets/scss/material.orcid.overwrites.scss/button.scss
leomendoza123: I can see this is just duplicated from the non-rtl CSS so you can just delete this line for now

src/assets/scss/material.orcid.overwrites.scss/button.scss
```diff
@@ -21,10 +21,41 @@ button.mat-button-wrap-text-with-icon {
...
+  margin-top: 12px;
+
+  a {
+    position: absolute;
+    right: 20px;
+    left: unset;
+    top: 5px;
+    img {
+      width: 24px;
```
leomendoza123, line 57: RTL is on top of the non-RTL CSS definition, so there is no need to repeat what stays the same, only the attributes that are overwritten. 

Meaning that you can remove all the attributes that are the same like:

```
width: 100%;
position: absolute;
top: 5px;
img {
      width: 24px;
```

normally only `left` and `right` related attributes are overwritten

src/assets/scss/material.orcid.overwrites.scss/button.scss
```diff
@@ -21,10 +21,41 @@ button.mat-button-wrap-text-with-icon {
...
+  }
+
+  .columns-4 & {
+    span {
+      display: inline-block;
+      width: 180px;
+      margin: auto;
+      word-break: break-word;
+      white-space: normal;
+      line-height: 24px;
```
leomendoza123, line 40: Not requesting a change here but just FYI: normally define font relate attributes like the `line-height` should be avoided so we do not overwrite font-related features from the app fonts.

src/assets/scss/material.orcid.overwrites.scss/button.scss
```diff
@@ -21,10 +21,41 @@ button.mat-button-wrap-text-with-icon {
   border: solid 1px rgba(0, 0, 0, 0.12);
   margin-top: 12px;
 
-  mat-icon {
+  a {
     position: absolute;
-    left: 5%;
-    top: 20%;
+    left: 20px;
+    top: 5px;
```
leomendoza123, line 27: image

This does not look in the middle on mobile

src/app/sign-in/components/social/social.component.html
DanielPalafox: I will remove it

src/assets/scss/material.orcid.overwrites.scss/button.scss
DanielPalafox: Ok

src/app/sign-in/components/social/social.component.html
DanielPalafox: I will erase it

src/assets/scss/material.orcid.overwrites.scss/button.scss
DanielPalafox: Ok ok. I will remote the other attributes

src/assets/scss/material.orcid.overwrites.scss/button.scss
DanielPalafox: Good catch!. I will fix it.

src/assets/scss/material.orcid.overwrites.scss/button.scss
DanielPalafox: Ok

PR #215, opened 2020-05-21
Remove extra input

leomendoza123:


leomendoza123:
[[QA] Recovery code field has double input line](https://trello.com/c/Y3PYBWki/6769-qa-recovery-code-field-has-double-input-line)

PR #216, opened 2020-05-21
Fix register steps navigation

leomendoza123:


leomendoza123:
[Register can't navigate from the first to the second step](https://trello.com/c/fkOxaCy4/6775-register-cant-navigate-from-the-first-to-the-second-step)

PR #218, opened 2020-05-22
Add visibility icons

leomendoza123:


leomendoza123:
[Register: Add visibility icons](https://trello.com/c/Q7sMADQg/6744-register-add-visibility-icons)

src/app/register/components/form-visibility/form-visibility.component.html
```diff
@@ -27,21 +27,30 @@ <h3 class="mat-subheading-2" i18n="@@register.visibility">
...
+          >Everyone (87% of users choose this)</ng-container
+        >
       </mat-radio-button>
       <mat-radio-button
         [value]="visibilityOptions[1]"
-        i18n="@@register.VisibilityTrusted"
+        class="orcid-radio-button-with-icons"
       >
-        Trusted parties (5% of users choose this)
+        <img src="./assets/vectors/visibility-trusted-source.svg" />
```
DanielPalafox, line 41: Attribute alt it is missing for accessibility in all the images

leomendoza123: So instead of adding alt labels, and since these icons are purely decorative icons I am setting those as hidden for accessible tools. 
In this way those do no get read by screen readers and create a confusing experience.

PR #221, opened 2020-05-23
Feature/6779 zendesk widget

leomendoza123:


leomendoza123:
[Zendesk widget is missing from orcid-angular pages](https://trello.com/c/T6s8oT2c/6779-zendesk-widget-is-missing-from-orcid-angular-pages)

PR #225, opened 2020-05-26
Add mat stepper docs

leomendoza123:


leomendoza123:
[Register: Add stepper component to storybook](https://trello.com/c/jvKjubua/6742-register-add-stepper-component-to-storybook)

PR #228, opened 2020-05-26
Fix a11y main content links

leomendoza123:


leomendoza123:
[[QA] "Skip to main content" doesn't work in angular pages](https://trello.com/c/Oj2smo2Q/6777-qa-skip-to-main-content-doesnt-work-in-angular-pages)

PR #229, opened 2020-05-26
Remove autocomplete on confirmation email

leomendoza123:


leomendoza123:
[[QA] Register: allows auto-complete in confirm primary email field](https://trello.com/c/SzyewB0N/6782-qa-register-allows-auto-complete-in-confirm-primary-email-field)

PR #231, opened 2020-05-26
Set qa captcha in testing mode

leomendoza123:


leomendoza123:
[QA do not have a testing captcha](https://trello.com/c/TkpSIoy6/6788-qa-do-not-have-a-testing-captcha)

PR #243, opened 2020-06-01
Fix a tag accessibility

DanielPalafox:
https://trello.com/c/CUUJ6Ybe/6795-qa-signin-links-above-authenticate-button-are-not-in-the-tab-order

### DIFF
--- a/src/app/sign-in/components/errors/deactivated/deactivated.component.html
+++ b/src/app/sign-in/components/errors/deactivated/deactivated.component.html
@@ -12,6 +12,7 @@
       If you can no longer access any emails associated with your iD, please
     </ng-container>
     <a
+      href
       class="mat-button-font"
       (click)="navigateTo('/help/contact-us')"
       i18n="@@ngOrcid.signin.contactSupport"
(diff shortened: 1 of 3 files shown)


DanielPalafox:
[[QA] Signin: links above authenticate button are not in the tab order](https://trello.com/c/CUUJ6Ybe/6795-qa-signin-links-above-authenticate-button-are-not-in-the-tab-order)

src/app/sign-in/components/errors/deactivated/deactivated.component.html
```diff
@@ -12,6 +12,7 @@
       If you can no longer access any emails associated with your iD, please
     </ng-container>
     <a
+      href
```
leomendoza123, line 15: This breaks the (click) event and redirects the user to the home page. 

To solve this you can import the `A11yLinkModule` into the Singin module like this https://github.com/ORCID/orcid-angular/blob/a20c2c09c7e3bcf9348ec258ae300d852b44c53a/src/app/register/register.module.ts#L68-L69, please test it as this module have not been used outside the register

PR #247, opened 2020-06-02
Focus first inputs on navigation to register and signin

leomendoza123:


leomendoza123:
[[QA] "Skip to main content" doesn't work in angular pages](https://trello.com/c/Oj2smo2Q/6777-qa-skip-to-main-content-doesnt-work-in-angular-pages)

src/app/register/components/form-personal/form-personal.component.html
```diff
@@ -93,6 +93,7 @@
         appTrim
         autocomplete="off"
         type="email"
+        #firstInput
```
DanielPalafox, line 96: This #firstInput is unnecessary since you already have it in givenNames

PR #251, opened 2020-06-02
Fix/6807 Add docs examples for push/pull from transifex

leomendoza123:


leomendoza123:
[Orcid Source translations migrator script](https://trello.com/c/lTvIXPli/6807-orcid-source-translations-migrator-script)

PR #254, opened 2020-06-03
Add require polyfill for ie11

leomendoza123:


leomendoza123:
[[QA] Register: add another email button doesn't work on IE11](https://trello.com/c/rotXjCzH/6813-qa-register-add-another-email-button-doesnt-work-on-ie11)

PR #255, opened 2020-06-03
Atemp to disable autocompletes on email confirmation

leomendoza123:
**Coping the same last message a put on the Trello card here:**

Unfortunately, disabling this is not really a supported feature on any major browser 
https://caniuse.com/#feat=input-autocomplete-onoff, and there are many open discussions about
 why and why not should browsers allow developers to disable autocompletes...

And there is a lot of info about why the "retype email" is bad UX...

But to avoid going too deep with this one, I would like to try one more thing that seems
 to be working for me.

### DIFF
--- a/src/app/register/components/form-personal/form-personal.component.html
+++ b/src/app/register/components/form-personal/form-personal.component.html
@@ -86,12 +86,15 @@
       <mat-label i18n="@@register.confirmEmail"
         >Confirm primary email</mat-label
       >
+      <!-- "confirm-email" is no really a valid autocomplete valid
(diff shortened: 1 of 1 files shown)


leomendoza123:
[[QA] Register: allows auto-complete in confirm primary email field](https://trello.com/c/SzyewB0N/6782-qa-register-allows-auto-complete-in-confirm-primary-email-field)

leomendoza123:
[[QA] Register: allows auto-complete in confirm primary email field](https://trello.com/c/SzyewB0N/6782-qa-register-allows-auto-complete-in-confirm-primary-email-field)

PR #259, opened 2020-06-03
Fix autocomplete value on register passwords

leomendoza123:


leomendoza123:
[Wrong browser autocomplete suggestions on register password fields](https://trello.com/c/68NoXVuK/6820-wrong-browser-autocomplete-suggestions-on-register-password-fields)

PR #266, opened 2020-06-05
Add mat button wrap

DanielPalafox:
https://trello.com/c/4vRV7MNY/6819-qa-signin-continue-as-account-name-button-cant-handle-long-names

### DIFF
--- a/src/assets/scss/material.orcid.overwrites.scss/button.scss
+++ b/src/assets/scss/material.orcid.overwrites.scss/button.scss
@@ -13,20 +13,7 @@ button.mat-button-wrap-text {
     max-width: 100%;
     text-align: center;
     white-space: normal;
-  }
-}
-
-button.mat-button-wrap-long-texts {
-  width: 100%;
-  margin-top: 12px;
-
-  span {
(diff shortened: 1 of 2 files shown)


DanielPalafox:
[[QA] Signin: continue as "account name" button can't handle long names](https://trello.com/c/4vRV7MNY/6819-qa-signin-continue-as-account-name-button-cant-handle-long-names)

src/assets/scss/material.orcid.overwrites.scss/button.scss
```diff
@@ -16,6 +16,20 @@ button.mat-button-wrap-text {
   }
 }
 
+button.mat-button-wrap-long-texts {
```
leomendoza123, line 19: What was the issue with the existing `mat-button-wrap-text`? 

It seems to me like it handles the long text properly 
image

DanielPalafox: The problem is that doesn't work for all scenarios.

image

But I will then refactor `mat-button-wrap-text` instead of creating a new one.

PR #277, opened 2020-06-09
Make matchValues validator case insensitive

leomendoza123:


leomendoza123:
[[QA] Register: "primary email" and "confirm primary email" shouldn't have to be in the same letter case](https://trello.com/c/MgK50P5t/6824-qa-register-primary-email-and-confirm-primary-email-shouldnt-have-to-be-in-the-same-letter-case)

PR #278, opened 2020-06-09
Use color for singin buttons from palettes, and use official seamless…

leomendoza123:
…access color


leomendoza123:
[[QA] Signin: a11y violation, insufficient color contrast on institutional signin button](https://trello.com/c/RB3vCzxO/6826-qa-signin-a11y-violation-insufficient-color-contrast-on-institutional-signin-button)

PR #280, opened 2020-06-09
Update register look

leomendoza123:


leomendoza123:
[Update visibility settings text on 'Register'](https://trello.com/c/rzmRNjRw/6815-update-visibility-settings-text-on-register)

src/app/register/components/form-visibility/form-visibility.component.html
```diff
@@ -73,4 +79,14 @@ <h3 class="mat-subheading-2" i18n="@@register.visibility">
       >Please choose a default visibility setting.</mat-error
     >
   </p>
+  <p>
```
DanielPalafox, line 82: p tag it is not necessary here

src/locale/properties/register/register.xx.properties
```diff
@@ -74,3 +74,6 @@ register.moreInfo=X
 register.securityAndNotifications=X
 register.visibilityAndTerm=X
 register.visibilityRequired=X
+register.VisibilityPrivate2=X
```
DanielPalafox, line 77: The properties `register.VisibilityPrivate1`, `register.VisibilityTrusted1` and `register.VisibilityEveryone1` are missing here and also in  the `rl` and `lr` properties files

src/app/register/components/form-visibility/form-visibility.component.html
amontenegro: Nice catch!!!

leomendoza123: is require for the margins 
image

PR #311, opened 2020-06-25
add username update in form sign in

DanielPalafox:
https://trello.com/c/LfCTI7Wu/6682-social-institutional-institutional-link-account-8


src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -43,6 +43,7 @@ export class FormSignInComponent implements OnInit {
   email: string
   orcidPrimaryDeprecated: string
   singInLocal = {} as SingInLocal
+  username: string
```
leomendoza123, line 46: I think this is never used

DanielPalafox: Great catch! I will remove it.

PR #312, opened 2020-06-26
fix real user orcid not found

DanielPalafox:


DanielPalafox:
[[QA] OAuth is broken](https://trello.com/c/GIvKIa7v/6849-qa-oauth-is-broken)

PR #314, opened 2020-07-01
When an email entered on the Registration screen is already in the db, auto-populate it when the user clicks to sign in

leomendoza123:
Fix username autofill from GET parameters on signin and add GET parameters to the `already exist` backend error on the register


leomendoza123:
[When an email entered on the Registration screen is already in the db, auto-populate it when the user clicks to sign in](https://trello.com/c/Rmi0WGbx/6854-when-an-email-entered-on-the-registration-screen-is-already-in-the-db-auto-populate-it-when-the-user-clicks-to-sign-in)

PR #318, opened 2020-07-07
add sign-in ga events

DanielPalafox:
https://trello.com/c/m9cYTx3V/6676-signin-ga-events

### DIFF
--- a/src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
+++ b/src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
@@ -199,22 +199,21 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
         this._gtag
           .reportEvent('RegGrowth', 'Sign-In', requestInfoForm)
           .subscribe(
-            () => {
-              this._router.navigate(['/oauth/authorize'], {
-                queryParams: {
-                  client_id: this.signInLocal.params.client_id,
-                  response_type: this.signInLocal.params.response_type,
-                  scope: this.signInLocal.params.scope,
-                  redirect_uri: this.signInLocal.params.redirect_uri,
-                },
-              })
-            },
-            () => {
-              // TODO @DanielPalafox display error using a toaster
-              console.log('Error oauth' + JSON.stringify(requestInfoForm.error))
-            }
+            () => this.oauthAuthorize(),
+            () => this.oauthAuthorize()
           )
       })
+  }
+
+  oauthAuthorize() {
+    this._router.navigate(['/oauth/authorize'], {
+      queryParams: {
+        client_id: this.signInLocal.params.client_id,
+        response_type: this.signInLocal.params.response_type,
+        scope: this.signInLocal.params.scope,
+        redirect_uri: this.signInLocal.params.redirect_uri,
+      },
+    })
   }
 
   updateUsername(email) {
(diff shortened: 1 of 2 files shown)


DanielPalafox:
[Signin: GA events](https://trello.com/c/m9cYTx3V/6676-signin-ga-events)

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -113,6 +115,10 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
           if (data.url.toLowerCase().includes('oauth/authorize')) {
             this.updateOauthSession(this.signInLocal.params)
           } else {
+            this._gtag.reportEvent('RegGrowth', 'Sign-In', 'Website').subscribe(
+              () => (this.window.location.href = data.url),
+              () => (this.window.location.href = data.url)
+            )
             this.navigateTo(data.url)
```
leomendoza123, line 122: So I think we won't require this navigate anymore, right?

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -191,6 +197,19 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
     this._oauthService
       .updateOauthSession(value)
       .subscribe((requestInfoForm) => {
+        this._gtag
+          .reportEvent(
+            'RegGrowth',
+            'Sign-In',
+            'OAuth ' +
```
leomendoza123, line 204: so take a look at this part of the service function 

https://github.com/ORCID/orcid-angular/blob/5623901e4951f1a768277f2b1f3d5518c0a6da0a/src/app/core/google-analytics/google-analytics.service.ts#L64-L65

you can just send just the `requestInfoForm` as the third parameter, so the Oauth string gets build on the service.

DanielPalafox: Ok

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
DanielPalafox: Yes, It is not required anymore. I will remove it.

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -191,14 +196,24 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
...
+                queryParams: {
+                  client_id: this.signInLocal.params.client_id,
+                  response_type: this.signInLocal.params.response_type,
+                  scope: this.signInLocal.params.scope,
+                  redirect_uri: this.signInLocal.params.redirect_uri,
+                },
+              })
+            },
+            () => {
+              // TODO @DanielPalafox display error using a toaster
```
leomendoza123, line 213: let's handle the error in the same way as the success for analytics events, as you do here 

https://github.com/ORCID/orcid-angular/blob/effa00ba6f3faa3c1d785bb4815fb6749b566e7d/src/app/sign-in/components/form-sign-in/form-sign-in.component.ts#L118-L120

this  since we want to keep things moving even if analytics is down

leomendoza123: I think we don't even want to show the toaster when analytics fails

DanielPalafox: Make sense. I will change it

PR #321, opened 2020-07-07
Fix/6863 qa blog feed not loading

leomendoza123:


leomendoza123:
[[QA] Blog feed not loading](https://trello.com/c/PARMrpVN/6863-qa-blog-feed-not-loading)

PR #322, opened 2020-07-07
Fix register issues

leomendoza123:


leomendoza123:
[[QA] Register in the new app is broken, can't get past step 1](https://trello.com/c/Izmap86o/6858-qa-register-in-the-new-app-is-broken-cant-get-past-step-1)

PR #464, opened 2020-07-21
Feature/6712 register oauth support 2

leomendoza123:


leomendoza123:
[Register - OAuth support](https://trello.com/c/ob0mVDFW/6712-register-oauth-support)

leomendoza123:
[Support Openid query parameters](https://trello.com/c/xapTqK4F/6875-support-openid-query-parameters)

leomendoza123:
[Register: GA events](https://trello.com/c/xoQd8mPc/6739-register-ga-events)

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -109,14 +111,14 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
       this.loading = true
 
       const isOauth = this.signInLocal.type === TypeSignIn.oauth
-      const $signIn = this._signIn.signIn(this.signInLocal, isOauth)
+      const $signIn = this._signIn.signIn(this.signInLocal, isOauth, isOauth)
```
DanielPalafox, line 114: I don't think this is right the parameters should be `updateUserSession` and `forceSessionUpdate` instead of isOauth

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -109,14 +111,14 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
...
       $signIn.subscribe((data) => {
         this.loading = false
         this.printError = false
 
         if (data.success) {
-          if (data.url.toLowerCase().includes('oauth/authorize')) {
-            this.updateOauthSession(this.signInLocal.params)
+          if (isARedirectToTheAuthorizationPage(data)) {
```
DanielPalafox, line 120: I think we should respect the camel case and it will be more readable to rename the method to `IsRedirectToAuthorizationPage` or something similar

DanielPalafox:
@leomendoza123 Great work Leo. I put some comments

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
leomendoza123: This is right (at the moment). Since only on Oauth mode logins we want to update the session, in force update mode. Here is more info about it https://github.com/ORCID/orcid-angular/blob/8abd9d6121dbbada43f5aafe54402d360ea61c6a/src/app/core/sign-in/sign-in.service.ts#L33-L34

leomendoza123: well in fact `forceSessionUpdate` is only needed when is an Oauth authorization for an Open id with a `forced login`. But it does not make any harm to use it for all oauts logins.

PR #466, opened 2020-07-29
Allow single char subdomains

leomendoza123:


leomendoza123:
[Unable to generate a reset password link using an email domain with 1 character](https://trello.com/c/UXYq9OLf/6895-unable-to-generate-a-reset-password-link-using-an-email-domain-with-1-character)

PR #467, opened 2020-07-31
Add shibboleth and social register support

leomendoza123:


leomendoza123:
[Register: shibboleth support](https://trello.com/c/W8XJDqsk/6844-register-shibboleth-support)

PR #474, opened 2020-08-18
Fix/6908 qa register fix error message for emails with deactivated and unclaimed accounts

leomendoza123:


src/app/register/components/backend-error/backend-error.component.html
```diff
@@ -22,4 +22,24 @@
...
+    >This email already exists in our system as an unclaimed record. Would you
+    like to</ng-container
+  >
+
+  <a (click)="navigateToSignin(value)" i18n="@@register.resendClaim"
+    >resend the claim email?</a
+  ><ng-container i18n="@@shared.questionMark">?</ng-container>
+</ng-container>
+
+<ng-container>
```
DanielPalafox, line 36: I think there is a `*ngIf` missing here

amontenegro:
[[QA] register - fix error message for emails with deactivated and unclaimed accounts](https://trello.com/c/nivDnc2z/6908-qa-register-fix-error-message-for-emails-with-deactivated-and-unclaimed-accounts)

PR #498, opened 2020-08-24
Feature/6927 speed upsidebar services

leomendoza123:


leomendoza123:
[[speed-up]Sidebar (services)](https://trello.com/c/qEN3s83p/6927-speed-upsidebar-services)

src/app/constants.ts
```diff
@@ -84,3 +84,255 @@ export const AMOUNT_OF_RETRIEVE_NOTIFICATIONS_PER_CALL = 10
...
+  'Saint Vincent and the Grenadines': 'VC',
+  Botswana: 'BW',
+  'Dominican Republic': 'DO',
+  Denmark: 'DK',
+  Uganda: 'UG',
+  Mexico: 'MX',
+  Suriname: 'SR',
+  Micronesia: 'FM',
+  'Saint Helena': 'SH',
+  Greenland: 'GL',
```
DanielPalafox, line 337: **OPTIONAL. I think it will better to store this in a file called `contries.json` and called with a service because it is adding almost 300 lines to `constants.ts`

leomendoza123: Will do it, that's a great idea

PR #499, opened 2020-08-24
Fix register backend error issues

leomendoza123:


leomendoza123:
[[QA] register - fix error message for emails with deactivated and unclaimed accounts](https://trello.com/c/nivDnc2z/6908-qa-register-fix-error-message-for-emails-with-deactivated-and-unclaimed-accounts)

PR #508, opened 2020-08-26
fix sign in Facebook

DanielPalafox:
https://trello.com/c/4qVPfo0V/6931-qa-sign-in-with-facebook-doesnt-work


DanielPalafox:
[[QA] Sign in with Facebook doesn't work](https://trello.com/c/4qVPfo0V/6931-qa-sign-in-with-facebook-doesnt-work)

PR #518, opened 2020-09-01
Filter out duplicated scope descriptions

leomendoza123:


leomendoza123:
[Display only one description for `/authenticate` or `openid` (not both)](https://trello.com/c/wEB760NU/6956-display-only-one-description-for-authenticate-or-openid-not-both)

DanielPalafox (approved):

PR #522, opened 2020-09-01
add 2fa page missing in institutional sign in

DanielPalafox:
https://trello.com/c/L1WNUVfT/6758-add-new-page-link-shibboleth-signin-for-new-institutional-linking-orcid-angular

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -98,6 +98,11 @@ const routes: Routes = [
       import('./search/search.module').then((m) => m.SearchModule),
   },
   {
+    path: ApplicationRoutes.twoFactor,
+    loadChildren: () =>
+      import('./two-factor/two-factor.module').then((m) => m.TwoFactorModule),
+  },
+  {
     path: '**',
     redirectTo: '/',
   },
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -99,6 +99,7 @@ const routes: Routes = [
   },
   {
     path: ApplicationRoutes.twoFactor,
+    canActivateChild: [LinkAccountGuard],
     loadChildren: () =>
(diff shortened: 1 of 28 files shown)


DanielPalafox:
[Add new page link /shibboleth/signin for new institutional linking orcid-angular](https://trello.com/c/L1WNUVfT/6758-add-new-page-link-shibboleth-signin-for-new-institutional-linking-orcid-angular)

src/app/two-factor/pages/two-factor/two-factor.component.ts
```diff
@@ -0,0 +1,83 @@
...
+          if (res.errors[0].includes('verification')) {
+            this.showBadVerificationCode = true
+          } else {
+            this.showBadRecoveryCode = true
+          }
+        } else if (res.redirectUrl) {
+          this.navigateTo(res.redirectUrl)
+        }
+      },
+      (error) => console.log(error)
```
leomendoza123, line 76: You are already using the ERROR_REPORT.STANDARD_VERBOSE to report this to the user and the console, so no need to handle the error here as well

src/app/app-routing.module.ts
```diff
@@ -97,6 +97,11 @@ const routes: Routes = [
     loadChildren: () =>
       import('./search/search.module').then((m) => m.SearchModule),
   },
+  {
+    path: ApplicationRoutes.twoFactor,
+    loadChildren: () =>
```
leomendoza123, line 102: A guard is required here, right? So no one can load this module without the required context

src/app/two-factor/pages/two-factor/two-factor.component.ts
DanielPalafox: Yes, you are right, great catch. I will remove it

leomendoza123 (approved):

PR #527, opened 2020-09-03
fix oauth issues with incorrect values

DanielPalafox:
https://trello.com/c/znWvsONw/6960-qa-oauth-issues-with-responses-to-calls-made-with-incorrect-clientids-and-scopes-or-unregistered-redirecturis


DanielPalafox:
[[QA] OAuth, issues with responses to calls made with incorrect client_ids and scopes, or unregistered redirect_uri's](https://trello.com/c/znWvsONw/6960-qa-oauth-issues-with-responses-to-calls-made-with-incorrect-clientids-and-scopes-or-unregistered-redirecturis)

src/app/types/request-info-form.endpoint.ts
```diff
@@ -20,7 +20,7 @@ export interface RequestInfoForm {
   nonce: string
   clientHavePersistentTokens: boolean
   scopesAsString: string
-  error: 'login_required' | 'interaction_required'
```
leomendoza123, line 23: are the more errors here than just 'login_required' | 'interaction_required', can't we just list those as well?

src/app/guards/authorize.guard.ts
```diff
@@ -54,13 +54,19 @@ export class AuthorizeGuard implements CanActivateChild {
     return this._user.getUserSession(queryParams).pipe(
       map((x) => x.oauthSession),
       map((session) => {
-        if (session.forceLogin || !oauthSessionUserIsLoggedIn(session)) {
-          return this.redirectToLoginPage(queryParams)
+        if (session.redirectUrl && session.responseType && session.redirectUrl.includes(session.responseType)) {
+          window.location.href = session.redirectUrl
```
leomendoza123, line 62: please use the window service

src/app/guards/authorize.guard.ts
```diff
@@ -54,13 +54,26 @@ export class AuthorizeGuard implements CanActivateChild {
     return this._user.getUserSession(queryParams).pipe(
       map((x) => x.oauthSession),
       map((session) => {
-        if (session.forceLogin || !oauthSessionUserIsLoggedIn(session)) {
-          return this.redirectToLoginPage(queryParams)
+        if (
+          session.redirectUrl &&
+          session.responseType &&
+          session.redirectUrl.includes(session.responseType)
```
leomendoza123, line 60: do we now all the possible values of `responseType`? Is it possible to add that to the types

leomendoza123 (changes requested):

src/app/guards/authorize.guard.ts
DanielPalafox: Ok, I will add them

src/app/types/request-info-form.endpoint.ts
DanielPalafox: Yes, there are more. Ok, I will list them all

src/app/types/request-info-form.endpoint.ts
```diff
@@ -31,3 +31,13 @@ export interface Scope {
   description: string
   longDescription: string
 }
+
+enum Errors {
```
leomendoza123, line 35: for this purpose, a type is required not an enum. 

Sinde the error itself will not be a enum

leomendoza123 (changes requested):

src/app/types/request-info-form.endpoint.ts
DanielPalafox: Yes, you are right. I will fix it

leomendoza123 (approved):

PR #534, opened 2020-09-08
fix register/sign form input red

DanielPalafox:
https://trello.com/c/FIqHqhej/6966-qa-clicking-register-now-on-signin-makes-username-field-red

### DIFF
--- a/src/app/shared/validators/username/username.validator.ts
+++ b/src/app/shared/validators/username/username.validator.ts
@@ -8,7 +8,7 @@ export class UsernameValidator {
     const orcidError = Validators.pattern(ORCID_REGEXP)(control)
     const orcidUriError = Validators.pattern(ORCID_URI_REGEXP)(control)
 
-    return control.value.length > 19
+    return control.value && control.value.length > 19
       ? validateUsername(orcidUriError, emailErrors)
       : validateUsername(orcidError, emailErrors)
   }
--- a/src/app/shared/validators/username/username.validator.ts
+++ b/src/app/shared/validators/username/username.validator.ts
@@ -4,11 +4,15 @@ import { OrcidValidators } from 'src/app/validators'
 
 export class UsernameValidator {
   static orcidOrEmail(control: AbstractControl): ValidationErrors | null {
+    if (control.value === null || control.value === '') {
+      return null
+    }
+
     const emailErrors = OrcidValidators.emailGeneric(control)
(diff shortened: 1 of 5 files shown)


DanielPalafox:
[[QA] clicking "register now" on /signin makes username field red](https://trello.com/c/FIqHqhej/6966-qa-clicking-register-now-on-signin-makes-username-field-red)

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -92,26 +91,29 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
...
 
   ngAfterViewInit(): void {
     this.firstInput.nativeElement.focus()
   }
 
   onSubmit() {
-    this.signInLocal.data = this.authorizationForm.getRawValue()
-
-    this.authorizationForm.markAllAsTouched()
+    this.addUsernameValidation()
```
leomendoza123, line 113: ** observation no changes require ** 

Addin the validation after submit is a intersting approach, I think I never see before. 
I believe it should work just fine.

an alternative to this is to write our own ErroStateMatcher as we have on the register. https://github.com/ORCID/orcid-angular/blob/master/src/app/register/ErrorStateMatcherForFormLevelErrors.ts

If we ever have any issue with this approach we can do try that.

leomendoza123 (changes requested):

src/app/shared/validators/username/username.validator.ts
```diff
@@ -8,7 +8,7 @@ export class UsernameValidator {
     const orcidError = Validators.pattern(ORCID_REGEXP)(control)
     const orcidUriError = Validators.pattern(ORCID_URI_REGEXP)(control)
 
-    return control.value.length > 19
+    return control.value && control.value.length > 19
```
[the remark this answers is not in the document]
DanielPalafox: Yes, you are right we should return something immediately there is no case to call `validateUsername` if it is null but we already have a validator for that which is `Validators.required`. So there is no case to return `invalidUsername`. I will do the fix

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
DanielPalafox: Ohh I see. Ok ok

leomendoza123 (approved):

PR #539, opened 2020-09-14
fix unnecessary first init.json call

DanielPalafox:
https://trello.com/c/PGW2b438/6887-speedup-oauth-custom-initjson-already-authorize-support


DanielPalafox:
[[speedup] /oauth/custom/init.json already authorize support](https://trello.com/c/PGW2b438/6887-speedup-oauth-custom-initjson-already-authorize-support)

DanielPalafox:
[[speedup] /oauth/custom/init.json already authorize support](https://trello.com/c/PGW2b438/6887-speedup-oauth-custom-initjson-already-authorize-support)

src/app/core/user/user.service.ts
```diff
@@ -142,9 +142,6 @@ export class UserService {
             this.loggingStateComesFromTheServer = true
             return this.userStatusHasChange(result)
           }),
-          // At the very beginning assumes the user is logged in,
-          // this is to avoid waiting for userStatus.json before calling userInfo.json and nameForm.json on the first load
-          startWith({ loggedIn: true, checkTrigger: -1 }),
```
leomendoza123, line 147: This is not great, because it removes the assumption of the user been logged in at the start, meaning that the nameForm and userInfo endpoints won't be called right at the start, but only after the user status endpoint responds. And making the load of that data wait is not great. So I think we can make this part of the code come back.

And I suggest here we can go into the user.service/ getOauthSession() and use the `checkTrigger`, and just call the init endpoint if check trigger is NOT a number bigger than -1. 

Since would want the session to be initialized right at the start (when the `checkTrigger === -1`),  when the user just log in (the `checkTrigger.postLoginUpdate === true`), and when for some reason a force refresh is call like when the user changed to use one of its delegated accounts (the `checkTrigger.forceSessionUpdate === true`).

leomendoza123 (changes requested):

src/app/core/user/user.service.ts
DanielPalafox: Ok

leomendoza123 (approved):

PR #541, opened 2020-09-16
Fix/6886 speedup oauth custom initjson error management

leomendoza123:


leomendoza123:
[[speedup] Open id promp=none does not return the 'interaction_required` error code](https://trello.com/c/vn59ESpm/6884-speedup-open-id-prompnone-does-not-return-the-interactionrequired-error-code)

DanielPalafox (approved):

PR #545, opened 2020-09-18
Remove confirmation email prefill

leomendoza123:


leomendoza123:
[[QA] confirm primary email field should never be pre-filled](https://trello.com/c/lNekHCAB/6985-qa-confirm-primary-email-field-should-never-be-pre-filled)

DanielPalafox (approved):

PR #546, opened 2020-09-18
Prefill existing Orcid ids on Oauth urls if exists at signin

leomendoza123:


leomendoza123:
[[QA] Authorization URL with ORCID iD that exists in the system should pre-fill the signin screen](https://trello.com/c/kXP2fSW5/6980-qa-authorization-url-with-orcid-id-that-exists-in-the-system-should-pre-fill-the-signin-screen)

DanielPalafox (approved):

PR #551, opened 2020-09-22
fix handle oauth session error

DanielPalafox:
https://trello.com/c/D8Zc4dCl/7002-qa-console-error-handleoauthsession

### DIFF
--- a/src/app/core/user/user.service.ts
+++ b/src/app/core/user/user.service.ts
@@ -280,6 +280,7 @@ export class UserService {
           if (
             typeof updateParameters.checkTrigger === `number` &&
             updateParameters.checkTrigger === -1
+            && updateParameters.loggedIn === false
           ) {
             return of(undefined)
           }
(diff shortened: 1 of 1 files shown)


DanielPalafox:
[[QA] console error handleOAuthSession](https://trello.com/c/D8Zc4dCl/7002-qa-console-error-handleoauthsession)

src/app/core/user/user.service.ts
```diff
@@ -280,6 +280,7 @@ export class UserService {
           if (
             typeof updateParameters.checkTrigger === `number` &&
             updateParameters.checkTrigger === -1
+            && updateParameters.loggedIn === false
```
leomendoza123: because of the nature of this first trigger

https://github.com/ORCID/orcid-angular/blob/0e0f6d37a3ca435e868c828cb3a69b7dfdbfb5bf/src/app/core/user/user.service.ts#L140

Will never be true. 

Can you please double-check, isn't this bringing back the original error?

leomendoza123 (changes requested):

src/app/core/user/user.service.ts
DanielPalafox: Yes, it is bring it back other error. I will fix it

leomendoza123:
Close in favor of #554

PR #557, opened 2020-09-23
fix institutional go back button

DanielPalafox:
https://trello.com/c/NN4EHT8w/6982-qa-institutional-linking-go

### DIFF
--- a/src/app/institutional/pages/institutional/institutional.component.scss
+++ b/src/app/institutional/pages/institutional/institutional.component.scss
@@ -4,7 +4,6 @@
 
 button {
   max-width: 113px;
-  border: solid 1px rgba(0, 0, 0, 0.12);
 }
 
 img {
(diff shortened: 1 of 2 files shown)


DanielPalafox:
[[QA] Institutional linking, go back button](https://trello.com/c/NN4EHT8w/6982-qa-institutional-linking-go-back-button)

src/app/institutional/pages/institutional/institutional.component.scss
```diff
@@ -23,6 +22,10 @@ img {
   display: flex;
   align-items: center;
   justify-content: space-between;
+
+  .continue {
+    border: solid 1px rgba(0, 0, 0, 0.12);
```
leomendoza123, line 27: please use the colors on a separate theme .scss

leomendoza123 (changes requested):

src/app/institutional/pages/institutional/institutional.component.scss
DanielPalafox: I was just reviewing the Register `next` button and doesn't have the border. I will remove it in the institutional so we will have both buttons with the same styles

leomendoza123 (approved):

PR #569, opened 2020-09-24
fix account linking and registration

DanielPalafox:
https://trello.com/c/JtrmQdPf/7018-qa-temp

### DIFF
--- a/src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
+++ b/src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
@@ -228,8 +228,7 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
         if (platform.social || platform.institutional) {
           if (this.signInData) {
             const { email, firstName, lastName, providerId } = this.signInData
-            this._user.getUserSession().
-            subscribe((userSession) => {
+            this._user.getUserSession().subscribe((userSession) => {
               const params = platform.queryParameters
               if (userSession.oauthSession) {
                 params['oauth'] = ''
@@ -240,7 +239,7 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
                   email,
                   firstName,
                   lastName,
-                  providerId
+                  providerId,
                 },
               })
             })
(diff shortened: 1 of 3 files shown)


src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -228,14 +228,20 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
...
             const { email, firstName, lastName, providerId } = this.signInData
-            this._router.navigate(['/register'], {
-              queryParams: {
-                ...platform.queryParameters,
-                email,
-                firstName,
-                lastName,
-                providerId,
-              },
+            this._user.getUserSession().subscribe((userSession) => {
```
leomendoza123, line 231: please add the first() pipe, so it close the subscription

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -228,14 +228,20 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
...
+              const params = platform.queryParameters
+              if (userSession.oauthSession) {
+                params['oauth'] = ''
```
leomendoza123, line 234: Just curious since I have a ticket related to this OAuth parameter. 

https://trello.com/c/QGZgx9ga/6981-qa-investigate-whether-oauth-can-be-removed-from-authorization-urls

Why do we need this for? I thought it was useless but I see you are adding it

leomendoza123 (changes requested):

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
DanielPalafox: Yes, it is used to know where to redirect if it must go to `oauth/authorize` but I will check later if we can remove it since it is not very useful or maybe we can put as a hidden parameter instead

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
DanielPalafox: ohh yeah. I will add it

leomendoza123 (approved):

DanielPalafox:
[[QA] OAuth, account linking + new registration](https://trello.com/c/JtrmQdPf/7018-qa-temp)

PR #583, opened 2020-09-29
fix oauth redirect to register if orcid id doesnt exist in db

DanielPalafox:
https://trello.com/c/qbSLbU6R/6979-qa-authorization-url-with-orcid-id-that-doesnt-exist-should-display-register-screen


DanielPalafox:
[[QA] Authorization URL with ORCID iD that doesn't exist should display register screen](https://trello.com/c/qbSLbU6R/6979-qa-authorization-url-with-orcid-id-that-doesnt-exist-should-display-register-screen)

leomendoza123 (approved):

PR #586, opened 2020-09-30
fix sign in and register when user refresh the page when it's an oaut…

DanielPalafox:
https://trello.com/c/ZLcE9ThM/7038-previusly-existing-oauth-applications-will-break-the-login

### DIFF
--- a/src/app/register/pages/register/register.component.ts
+++ b/src/app/register/pages/register/register.component.ts
@@ -102,7 +102,10 @@ export class RegisterComponent implements OnInit, AfterViewInit {
             this.FormGroupStepA = this.prefillRegisterForm(
               this.platform.queryParameters
             )
-          } else if (session.oauthSession && Object.keys(platform.queryParameters).length !== 0) {
+          } else if (
+            session.oauthSession &&
+            Object.keys(platform.queryParameters).length !== 0
+          ) {
             this.requestInfoForm = session.oauthSession
             this.FormGroupStepA = this.prefillRegisterForm(this.requestInfoForm)
           }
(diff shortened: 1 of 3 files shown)


src/app/register/pages/register/register.component.ts
```diff
@@ -102,7 +102,7 @@ export class RegisterComponent implements OnInit, AfterViewInit {
             this.FormGroupStepA = this.prefillRegisterForm(
               this.platform.queryParameters
             )
-          } else if (session.oauthSession) {
+          } else if (session.oauthSession && Object.keys(platform.queryParameters).length !== 0) {
```
leomendoza123: Is this making you have an error?
Because if there is a oauthSession, the form can be prefilled with no problem even when there is no parameters

DanielPalafox:
[Previusly existing Oauth applications will break the login](https://trello.com/c/ZLcE9ThM/7038-previusly-existing-oauth-applications-will-break-the-login)

src/app/sign-in/components/form-sign-in/form-sign-in.component.ts
```diff
@@ -82,8 +82,7 @@ export class FormSignInComponent implements OnInit, AfterViewInit {
         session = session as UserSession
         platform = platform as PlatformInfo
         this.platform = platform
-
-        if (session.oauthSession) {
+        if (session.oauthSession && this.isQueryParametersEmpty(this.platform.queryParameters)) {
```
leomendoza123: a- seems like this function should be called `isQueryParameters*Not*Empty`

b- even when the query parameters are empty the Oauth session should be loaded to make this one possible https://trello.com/c/mwId6nLd/7036-the-oauth-session-is-forgotten

leomendoza123 (changes requested):

DanielPalafox:
[In BibTeX citation don't include the authors name twice if they are listed as a contributor (3)](https://trello.com/c/ixA5GC5r/6879-in-bibtex-citation-dont-include-the-authors-name-twice-if-they-are-listed-as-a-contributor-3)

PR #592, opened 2020-10-07
fix response type as parameter

DanielPalafox:
Fix response type as a parameter in the redirect url


amontenegro:
https://trello.com/c/6JVpa3Tl/7053-allow-the-word-code-in-the-redirect-uri

amontenegro (approved):

PR #593, opened 2020-10-07
Move all assets from the build to a single shared folder

leomendoza123:
**Fix attempt** 

- Making available the .JS chunks at all times.  Avoiding app breaks on weird scenarios where a .js file for a certain language X is called while having a cookie language Y.

**Important note** 

- The ideal or expected behavior is that when a language is changed in any tab all the other tabs will be reloaded on the recently selected language. At the moment this won't happen but the application won't get broken anymore when the user interacts with a previous tab. 

- This will also highly likely solve scenarios like the one reported by Paula since all the .JS files will be always available, making less likely that a .JS file returning a redirect to the index.html.

**Possible follow-ups to discuss** 

- catch and report errors that happen before angular loads (like the one reported by Paula)
- catch,  report, clear cache, and reload when a  [chuck error loads](https://medium.com/@kamrankhatti/angular-lazy-routes-loading-chunk-failed-42b16c22a377) happens
- Complete disable the index.html cache using `no-store`, to avoid unexpected issues with Cloudflare removing the etags
- explicitly set  [nginx configuration](https://www.alex-arriaga.com/failed-to-load-module-script-the-server-responded-with-a-non-javascript-mime-type-of-text-html-in-nginx-when-deploying-an-angular-application/) to return JS files with javascript headers. This has been an issue reported by many on [angular github](https://github.com/angular/angular/issues/30835)

**How to test**

- Make Cloudflare remove etags
- change languages on different tabs


DanielPalafox (approved):

leomendoza123:
[Multiple tabs might break the language change](https://trello.com/c/jt2mBvMy/7051-multiple-tabs-might-break-the-language-change)

PR #603, opened 2020-10-22
fix OAuth request, the register is shown when the prompt login parame…

DanielPalafox:
https://trello.com/c/qrsCFitY/7092-the-register-is-shown-on-promptlogin-requests-for-logged-in-users

### DIFF
--- a/src/app/core/user/user.service.ts
+++ b/src/app/core/user/user.service.ts
@@ -184,9 +184,9 @@ export class UserService {
           orcidUrl: this.getOrcidUrl(data),
           effectiveOrcidUrl: this.getOrcidUrl(data, true),
           oauthSessionIsLoggedIn:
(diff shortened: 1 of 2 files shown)


leomendoza123 (changes requested):
I see how this is better because it now matches the type definition. 
But looking into the code, I don't believe this is making any functional difference, can you please double-check it.

DanielPalafox:
[The register is shown on prompt=login requests (for logged in users)](https://trello.com/c/qrsCFitY/7092-the-register-is-shown-on-promptlogin-requests-for-logged-in-users)

DanielPalafox:
Ohh my bad one commit is missing

amontenegro (approved):

leomendoza123 (approved):

PR #614, opened 2020-11-02
Fix/reduce server status calls

leomendoza123:


leomendoza123:
[The new app is doing too many .JSON calls compared to the old app](https://trello.com/c/CVnAlo0L/7080-the-new-app-is-doing-too-many-json-calls-compared-to-the-old-app)

amontenegro (approved):

PR #612, opened 2020-11-02
Fix/test oauth request with reigstered emails

leomendoza123:
Please review https://github.com/ORCID/orcid-angular/pull/611 first since this PR also includes those commits

image

PR #617, opened 2020-11-03
Feature/remove oauth parameters from url

leomendoza123:
Include #612 changes
image

### DIFF
--- a/src/app/sign-in/pages/sign-in/sign-in.component.html
+++ b/src/app/sign-in/pages/sign-in/sign-in.component.html
@@ -15,7 +15,6 @@
               aria-level="1"
               i18n="@@ngOrcid.signin.title"
               *ngIf="!isOauthError"
-              id="signing-button"
             >
               Sign in
             </mat-card-title>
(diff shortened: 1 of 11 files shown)


leomendoza123:
[[QA] investigate whether "&oauth=" can be removed from authorization URLs](https://trello.com/c/QGZgx9ga/6981-qa-investigate-whether-oauth-can-be-removed-from-authorization-urls)

src/app/sign-in/pages/sign-in/sign-in.component.html
```diff
@@ -42,6 +41,7 @@
                   [disabled]="loading"
                   *ngIf="!show2FA"
                   i18n="@@ngOrcid.signin.button"
+                  id="signing-button"
```
DanielPalafox, line 44: I think it should be `signin-button` instead

DanielPalafox (changes requested):

src/app/sign-in/pages/sign-in/sign-in.component.html
leomendoza123: Fixing on #618

DanielPalafox (approved):

PR #620, opened 2020-11-04
fix url that are redirected to the oauth

DanielPalafox:
https://trello.com/c/gtZWRa4W/7107-issue-with-urls-that-are-redirect-to-the-oauth-page

### DIFF
--- a/src/app/constants.ts
+++ b/src/app/constants.ts
@@ -1,5 +1,3 @@
-import { RequestInfoForm } from './types'
-
 export { COUNTRY_NAMES_TO_COUNTRY_CODES } from './constants-country-codes'
 
 // Custom email REGEXP
--- a/src/app/constants.ts
+++ b/src/app/constants.ts
@@ -1,3 +1,5 @@
+import { RequestInfoForm } from './types'
+
 export { COUNTRY_NAMES_TO_COUNTRY_CODES } from './constants-country-codes'
 
 // Custom email REGEXP
@@ -105,4 +107,8 @@ export function objectToUrlParameters(object: Object) {
   return Object.keys(object)
     .map((key) => `${key}=${encodeURIComponent(object[key])}`)
     .join('&')
+}
+
+export function buildClientString(requestInfoForm: RequestInfoForm) {
+  return requestInfoForm.memberName + ' - ' + requestInfoForm.clientName
 }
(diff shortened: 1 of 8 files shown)


DanielPalafox:
[Issue with URLs that are redirect to the Oauth page](https://trello.com/c/gtZWRa4W/7107-issue-with-urls-that-are-redirect-to-the-oauth-page)

src/app/constants.ts
```diff
@@ -106,3 +108,7 @@ export function objectToUrlParameters(object: Object) {
     .map((key) => `${key}=${encodeURIComponent(object[key])}`)
     .join('&')
 }
+
+export function buildClientString(requestInfoForm: RequestInfoForm) {
```
leomendoza123, line 112: Please do not duplicate the code we already have on `google-analytics.service`... we can keep the same `this.requestInfoForm || 'Website'` when calling  `reportEvent` since report event already does build the client string when it gets a requestInfoForm

src/app/register/pages/register/register.component.ts
```diff
@@ -152,9 +155,11 @@ export class RegisterComponent implements OnInit, AfterViewInit {
           if (response.url) {
             this._gtag
               .reportEvent(
-                'RegGrowth',
                 'New-Registration',
-                this.requestInfoForm || 'Website'
+                'RegGrowth',
```
leomendoza123, line 159: Great catch! the ReGrowth is a category NOT an action. 
Now this will break my wrong tests here:

https://github.com/ORCID/orcid-angular/blob/88ab2416b78ba3a6830a6042798c7eda3430dc17/cypress/support/google-analytics.commands.js#L103-L105

So I will add one or two commits to this PR to fix it

src/app/constants.ts
leomendoza123: 🚨I will change this myself since I have to fix the e2e any way

leomendoza123 (changes requested):

leomendoza123:
All tests passing (including two new tests) 
image

leomendoza123:
All tests passing

leomendoza123 (approved):

leomendoza123:
[Google Analytics Categories and Actions is been wrongly send](https://trello.com/c/wa3UkLDy/7105-google-analytics-categories-and-actions-is-been-wrongly-send)

PR #627, opened 2020-11-10
Add parameters when moving back to the login

leomendoza123:


DanielPalafox (approved):

DanielPalafox:
[Oauth + Account Linking process is broken](https://trello.com/c/DIeeZy72/7129-oauth-account-linking-process-is-broken)

PR #628, opened 2020-11-10
Allow Oauth and social/platform types at the same time

leomendoza123:


leomendoza123:
[Oauth + Account Linking process is broken](https://trello.com/c/DIeeZy72/7129-oauth-account-linking-process-is-broken)

amontenegro (approved):

PR #631, opened 2020-11-19
fix removing parameters in institutional/social oauth

DanielPalafox:
https://trello.com/c/EiZOE6b1/7138-oauth-urls-with-email-a-cancelled-account-link-registrations-will-remove-the-email-parameter-from-the-oauth-url

### DIFF
--- a/src/app/link-account/pages/link-account/link-account.component.ts
+++ b/src/app/link-account/pages/link-account/link-account.component.ts
@@ -95,9 +95,9 @@ export class LinkAccountComponent implements OnInit {
         this._router.navigate([ApplicationRoutes.signin], {
           queryParams: {
             ...params,
-            email: (params.email ? params.email : undefined),
-            firstName: (params.firstName ? params.firstName : undefined),
-            lastName: (params.lastName ? params.lastName : undefined),
+            email: params.email ? params.email : undefined,
+            firstName: params.firstName ? params.firstName : undefined,
+            lastName: params.lastName ? params.lastName : undefined,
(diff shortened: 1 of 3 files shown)


DanielPalafox:
[Oauth URLs with email + a cancelled account-link registrations, will remove the email parameter from the Oauth URL](https://trello.com/c/EiZOE6b1/7138-oauth-urls-with-email-a-cancelled-account-link-registrations-will-remove-the-email-parameter-from-the-oauth-url)

leomendoza123 (changes requested):
This change won't really prevent the OAuth parameters from been remove. This will only change the Oauth URL from having the original email to have the email given by the login facebook/google or institution.

That will be weird because if for instance, a university sends the user with a OAuth URL like:

`authorize?...clientId=App-1&email=Dani@universityX.org&name=daniel`

this might end up with a Url like 

`authorize?...clientId=App-1&email=daniel@gmail.org&name=Dani`


And this will be recorded in our GA data. 

A real solution is sending the shib/social data through the user session to fill the register, instead of using query parameters. But since this is a refactor that will affect multiple parts of the app we might rather wait to have more automation before going with this.

amontenegro:
@DanielPalafox @leomendoza123 lets revisit this, I will send you an invite for next week so we can discuss whats the better way to solve this problem.

leomendoza123:
@DanielPalafox @amontenegro I propose closing this one in favor of https://github.com/ORCID/orcid-angular/pull/724

PR #647, opened 2020-11-25
fix inaccurate notification count

DanielPalafox:
https://trello.com/c/e9Ak2jOK/7171-inaccurate-notification-count


DanielPalafox:
[Inaccurate notification count](https://trello.com/c/e9Ak2jOK/7171-inaccurate-notification-count)

PR #655, opened 2020-12-08
Feature/7072 modify sign in event to ga

leomendoza123:


leomendoza123:
[Modify SIGN-IN event to GA](https://trello.com/c/YsTvisp2/7072-modify-sign-in-event-to-ga)

DanielPalafox (approved):

PR #656, opened 2020-12-09
fix oauth with the word social in the redirect uri

DanielPalafox:
https://trello.com/c/rf8ziw3a/7184-critical-redirect-uris-with-the-word-social-result-in-a-call-to-the-wrong-endpoint

### DIFF
--- a/src/app/cdk/platform-info/platform-info.service.ts
+++ b/src/app/cdk/platform-info/platform-info.service.ts
@@ -213,10 +213,16 @@ export class PlatformInfoService {
     this.platform = {
       ...this.platform,
       hasOauthParameters: this.hasOauthParameters(),
-      social: this.window.location.pathname.toLowerCase().indexOf('social-linking') >= 0,
+      social:
+        this.window.location.pathname.toLowerCase().indexOf('social-linking') >=
+        0,
       institutional:
-        this.window.location.pathname.toLowerCase().indexOf('institutional-linking') >= 0 ||
-        this.window.location.pathname.toLowerCase().indexOf('institutional-signin') >= 0,
+        this.window.location.pathname
+          .toLowerCase()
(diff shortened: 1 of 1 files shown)


DanielPalafox:
[[Critical] Redirect URIs with the word "social" result in a call to the wrong endpoint](https://trello.com/c/rf8ziw3a/7184-critical-redirect-uris-with-the-word-social-result-in-a-call-to-the-wrong-endpoint)

src/app/cdk/platform-info/platform-info.service.ts
```diff
@@ -213,9 +213,16 @@ export class PlatformInfoService {
...
+        this.window.location.pathname.toLowerCase().indexOf('social-linking') >=
+        0,
       institutional:
-        this.window.location.href.toLowerCase().indexOf('institutional') >= 0,
+        this.window.location.pathname
+          .toLowerCase()
+          .indexOf('institutional-linking') >= 0 ||
+        this.window.location.pathname
+          .toLowerCase()
+          .indexOf('institutional-signin') >= 0,
```
leomendoza123, line 225: Nice fix! 
Maybe we should also use window.location.pathname here as well 

https://github.com/ORCID/orcid-angular/blob/bde83b34b66386012920018cc445e0573c83e1d8/src/app/cdk/platform-info/platform-info.service.ts#L179

So the app don't get break even if the user sends `social-linking` at any point on the query parameters

leomendoza123 (changes requested):

src/app/cdk/platform-info/platform-info.service.ts
DanielPalafox: Done!

leomendoza123 (approved):

PR #668, opened 2021-01-14
Update zendesk autofill

auumgn:
Re-opening to make minor changes to the autofill


src/app/core/zendesk/zendesk.service.ts
```diff
@@ -80,7 +92,7 @@ export class ZendeskService {
               },
             },
             { id: 'subject', prefill: { '*': subject || '' } },
-
+            { id: '360039623413', prefill: { '*': uri || '' } },
```
leomendoza123, line 95: URI is an array, we might cast it as a string with a `''+ uri`

leomendoza123 (changes requested):
Great PR @auumgn!
I have added two comments about  it

leomendoza123 (approved):

auumgn:
[Auto populate Zendesk ticket OAuth errors (5)](https://trello.com/c/5NdhoT84/7084-auto-populate-zendesk-ticket-oauth-errors-5)

PR #675, opened 2021-02-01
Catch and handles errors caused by Google Analytics

leomendoza123:


leomendoza123:
[[QA] gtag call time-out error during OAuth](https://trello.com/c/9iRFVusb/7225-qa-gtag-call-time-out-error-during-oauth)

amontenegro (approved):

PR #693, opened 2021-02-23
add cypress websites testing

DanielPalafox:


cypress/integration/my-orcid/my-orcid-side-bar.e2e.js
```diff
@@ -70,21 +73,158 @@ describe.only('My Orcid sidebar' + runInfo(), () => {
...
+        .within(() => {
+          cy.get('#edit-button').click()
+        })
+        .get('#modal-container')
+        .get('#add-link')
+        .click()
+        .get('.mat-form-field-flex')
+        .eq(0)
+        .click({ multiple: true })
+        .type('Description Website 1', { delay: 0 })
```
leomendoza123, line 146: The `multiple: true ` is not require since the `.eq(0)` should return only one element

leomendoza123 (changes requested):

cypress/integration/my-orcid/my-orcid-side-bar.e2e.js
leomendoza123: this is not a blocking issue

leomendoza123 (approved):

PR #696, opened 2021-02-24
fix focus, margin websites and other names

DanielPalafox:


src/app/cdk/side-bar/modals/modal-websites/modal-websites.component.ts
```diff
@@ -99,7 +100,7 @@ export class ModalWebsitesComponent implements OnInit, OnDestroy {
       group[website.putCode] = new FormGroup({
         description: new FormControl(website.urlName),
         url: new FormControl(website.url.value, {
-          validators: [Validators.pattern(URL_REGEXP)],
+          validators: [Validators.required, Validators.pattern(URL_REGEXP)],
```
leomendoza123, line 103: Our `URL_REGEXP` does allow URLS like `google.com` the backend requires a complete email. 
We need to either add a backend async validator or add a new frontend local validation

leomendoza123 (changes requested):
Seems like the new other names inputs are not autofocus at the moment

leomendoza123 (approved):

PR #699, opened 2021-02-24
Add affiliations test descriptions

leomendoza123:


cypress/integration/my-orcid/my-orcid-affiliations.e2e.js
```diff
@@ -0,0 +1,53 @@
...
+      it('show the item source', () => {})
+      it('show the item "make a copy and edit" button', () => {})
+      it('can be deleted', () => {})
+    })
+
+    describe('display a user with an item added with multiple sources', () => {
+      it('show the number of items on the title as 1', () => {})
+      it('show the item multiple sources', () => {})
+      it('expand and show the multiple sources of the same item', () => {
+        // TODO ??
```
leomendoza123, line 31: @dandineen At the moment there are not designed to expand affiliations sources. 
I am planing to use and implement [like the one display for works](zpl.io/a3LAkD1)

But removing the checkboxes 
image

cypress/integration/my-orcid/my-orcid-affiliations.e2e.js
```diff
@@ -0,0 +1,53 @@
...
+      })
+    })
+
+    describe('display a user with many items', () => {
+      // TODO ??
+      it('show the number of items on the title as N', () => {})
+      it('can sort the items by the end date', () => {})
+      it('can sort the items by start date', () => {})
+      it('can sort the items by start title', () => {})
+    })
```
leomendoza123, line 41: @dandineen The sort menu that best matches the one on production affiliations is shown for works displays

image

But it only shows order by `start date`, `end date and `alphabetically`, and not the `start date reverse`, `end date reverse` and  `alphabetically`.  Is the reverse something we want to remove?

leomendoza123: I am planning to add a sort arrow like the one use [Here](https://material.angular.io/components/sort/overview) on the Table with sorting example 

image

image

cypress/integration/my-orcid/my-orcid-affiliations.e2e.js
dandineen: @leomendoza123 That's exactly what I was thinking should happen :-) The overall behaviour should be identical to works but you can't select multiple items.

cypress/integration/my-orcid/my-orcid-affiliations.e2e.js
dandineen: We don't currently give the `reverse` options anywhere do we?

Oh, hold on! If you click the same sort filter a second time it reverses the sorting direction. The icons are so small I didn't realise they were changing!

I think the _behaviour_ should stay the same but use those larger, clearer and more obvious arrow icons as you suggested above.

PR #747, opened 2021-04-08
add reactivation

DanielPalafox:


src/app/cdk/platform-info/platform-info.service.ts
```diff
@@ -258,4 +265,9 @@ export class PlatformInfoService {
   public getCurrentRoute(): string {
     return this._router.url
   }
+
+  getReactivationCode(): string {
+    const segments = window.location.href.split('/');
```
leomendoza123: let's not use the window directly.

leomendoza123 (approved):

PR #751, opened 2021-04-09
Fix/7312 us data transfer checkbox on reactivation

DanielPalafox:


DanielPalafox:
[US data-transfer checkbox on reactivation (3)](https://trello.com/c/304nQ54c/7312-us-data-transfer-checkbox-on-reactivation-3)

PR #756, opened 2021-04-14
fix styles in my orcid

DanielPalafox:


src/app/record/components/work-stack/work-stack.component.scss
```diff
@@ -17,3 +17,7 @@
     position: relative;
   }
 }
+
+a {
```
leomendoza123, line 21: Should we maybe create a global style with this?  here https://github.com/ORCID/orcid-angular/blob/dev/src/styles.scss#L59-L69

src/app/record/components/funding/funding.component.html
```diff
@@ -2,12 +2,12 @@
...
 <!-- GENERAL DATA DISPLAY BY DEFAULT -->
 <app-panel-data
-  class="defaul-data-loaded orc-font-body-small"
+  class="default-data-loaded orc-font-body-small"
   [borderBottom]="true"
 >
   <div>
     <div class="general-data">
-      <ng-container *ngIf="funding.startDate.year"
+      <ng-container *ngIf="funding && funding.startDate"
```
leomendoza123, line 10: We might use conditional chaining for a cleaner look. 
But this is fine, no need to change

leomendoza123 (changes requested):
Just add the URL underline as a global class

src/app/record/components/work-stack/work-stack.component.scss
DanielPalafox: Done!

leomendoza123 (approved):

PR #796, opened 2021-05-12
fix expand activities buttons, my public afiliation error, etc

DanielPalafox:


src/app/core/record-affiliations/record-affiliations.service.ts
```diff
@@ -88,6 +88,28 @@ export class RecordAffiliationService {
     type,
     options?: UserRecordOptions
   ): Observable<Affiliation> {
+    console.log(JSON.stringify(options))
```
leomendoza123: log

src/app/core/record-affiliations/record-affiliations.service.ts
```diff
@@ -103,6 +103,26 @@ export class RecordAffiliationService {
       )
   }
 
+  private getPublicAffiliationDetails(
```
leomendoza123, line 106: Seems like nothing is calling this

leomendoza123 (changes requested):

src/app/core/record-affiliations/record-affiliations.service.ts
DanielPalafox: You are right. I am gonna to remove it

leomendoza123 (approved):

PR #803, opened 2021-05-17
Boost search results in the UI

amontenegro:


amontenegro:
[Boost records with affiliations in search results (which pushes spam down) [Private boards only] (3)](https://trello.com/c/VPSJWUDc/8-boost-records-with-affiliations-in-search-results-which-pushes-spam-down-private-boards-only-3)

leomendoza123 (approved):

PR #871, opened 2021-05-25
Fix/batch fixes

leomendoza123:


src/locale/properties/record/record.rl.properties
```diff
@@ -31,3 +31,8 @@ record.translatedTitle=RL
 record.contributors=RL
 record.countryPublication=RL
 record.description=RL
+record.noPublicInfo=RL
+record.lockedRecord=RL
+thisAccountHasBeenDeprecated2=RL
+userInfo.PRIMARY_RECORD=RL
```
DanielPalafox, line 37: could you rename the property and use camelcase instead of the underscore

src/locale/properties/record/record.en.properties
```diff
@@ -29,3 +29,7 @@ record.editThisRecord=Edit this record
 record.ogDescription=ORCID provides an identifier for individuals to use with their name as they engage in research, scholarship, and innovation activities.
 record.countryPublication=Country of publication
 record.description=Description
+record.noPublicInfo=No public information available.
+record.lockedRecord=This ORCID Record is locked
+thisAccountHasBeenDeprecated=This account has been deprecated, please see account
```
DanielPalafox, line 34: could you replace the property and use the same format like `record.accountDeprecated` or something similar

DanielPalafox (changes requested):

DanielPalafox (approved):

PR #890, opened 2021-05-27
adding not found page view

### LINKED ISSUE #889: GENERAL - Show an appropriated 404 message when a user does not exist (Leo)

DanielPalafox:
Fix #889


src/styles.scss
```diff
@@ -33,7 +33,7 @@ app-root {
   flex-direction: column;
 }
 .router-container {
-  padding-bottom: 140px; //remove footer height
+  padding-bottom: 240px; //remove footer height
```
leomendoza123, line 36: Is this ok? I am afraid my break  the home page and other pages

leomendoza123 (changes requested):

src/styles.scss
DanielPalafox: Yes, actually it is a rollback

PR #909, opened 2021-06-07
Feature/sorting

leomendoza123:


src/app/core/record-fundings/record-fundings.service.ts
```diff
@@ -76,11 +82,22 @@ export class RecordFundingsService {
...
+
+  private getAndSortFundings(
+    options: UserRecordOptions
+  ): Observable<FundingGroup[]> {
     return this._http
       .get<FundingGroup[]>(
         environment.API_WEB +
-          `fundings/fundingGroups.json?sort=date&sortAsc=false`
+          `fundings/fundingGroups.json?` +
+          '/fundingGroups.json?' +
```
DanielPalafox, line 96: I dont think this line it should be here

src/app/core/record-works/record-works.service.ts
```diff
@@ -92,7 +92,7 @@ export class RecordWorksService {
             '&sort=' +
             (options.sort != null ? options.sort : true) +
```
DanielPalafox, line 93: sort by date it is missing here

src/locale/properties/shared/shared.en.properties
```diff
@@ -77,3 +77,9 @@ shared.sortResearch=Sort Research Resources
 shared.addPeerReviews=Add Peer Review
 shared.sortPeerReviews=Sort Peer Reviews
 share.loadMore=Load more
+share.sortTitle=Title
```
DanielPalafox, line 80: Translations for the other languages are missing, we have the translations for these words in ORCID-Source

src/app/record/pages/my-orcid/my-orcid.component.ts
```diff
@@ -66,6 +63,7 @@ export class MyOrcidComponent implements OnInit, OnDestroy {
       .subscribe((userRecord) => {
         this.recordWithIssues = userRecord?.userInfo?.RECORD_WITH_ISSUES
         this.userRecord = userRecord
+        this.loading = true
```
DanielPalafox, line 66: The PR https://github.com/ORCID/ORCID-Source/pull/6241 it is necessary to have the PR merged since the current code is throwing a 500 and the server won't reply and `loading` won't change to true when a user is not found

DanielPalafox (changes requested):

src/app/record/pages/my-orcid/my-orcid.component.ts
leomendoza123: we are going to merge this on the next QA release, after the PROD release according to angel

PR #912, opened 2021-06-07
implement printable version button

DanielPalafox:


DanielPalafox:
[[speedup Q2] Implement printable version button](https://trello.com/c/w9diRsF5/7273-speedup-q2-implement-printable-version-button)

PR #914, opened 2021-06-16
Fix/7254 upgrade angular so drag and drop works on rtl languages 3

DanielPalafox:


DanielPalafox:
[Upgrade Angular so drag and drop works on RTL languages (3)](https://trello.com/c/evxMDkjv/7254-upgrade-angular-so-drag-and-drop-works-on-rtl-languages-3)

leomendoza123 (approved):

PR #915, opened 2021-06-16
fix print redirect

DanielPalafox:


DanielPalafox:
[[QA][public page] printable version doesn't work](https://trello.com/c/MZr4lYLG/7370-qapublic-page-printable-version-doesnt-work)

leomendoza123 (approved):

PR #919, opened 2021-06-17
fix panels name if there is nothing to display

DanielPalafox:


DanielPalafox:
[[QA][public page] Hide name panel when name is set to private or trusted parties](https://trello.com/c/mU1vvGQr/7375-qapublic-page-hide-name-panel-when-name-is-set-to-private-or-trusted-parties)

leomendoza123 (approved):

PR #920, opened 2021-06-17
fix resize names panel

DanielPalafox:


DanielPalafox:
[[QA][public page] Don't resize name panels when one of them is not displayed](https://trello.com/c/QYyI7RCu/7376-qapublic-page-dont-resize-name-panels-when-one-of-them-is-not-displayed)

leomendoza123 (approved):

PR #921, opened 2021-06-17
fix source date in also know as

DanielPalafox:


DanielPalafox:
[[QA][public page] Dates not shown in also known as panel](https://trello.com/c/Un3sYFpT/7369-qapublic-page-dates-not-shown-in-also-known-as-panel)

leomendoza123 (approved):

DanielPalafox:
[[QA][public page] When name and published name are not publicly visible, the source of other names should be the ORCID iD](https://trello.com/c/LM1kbFKs/7371-qapublic-page-when-name-and-published-name-are-not-publicly-visible-the-source-of-other-names-should-be-the-orcid-id)

PR #923, opened 2021-06-18
fix other names comma

DanielPalafox:


DanielPalafox:
[[QA][public page] Issues with name panels](https://trello.com/c/dEK3hIMi/7368-qapublic-page-issues-with-name-panels)

src/app/record/components/top-bar/top-bar.component.scss
```diff
@@ -103,6 +103,14 @@ app-top-bar-actions {
       .other-names-wrapper {
         padding: 16px;
 
+        p {
+          margin: 0px;
+        }
+
+        .comma {
+          margin-left: -0.3em;
```
leomendoza123, line 111: I think this won't be working very well on RTL languages. 
Should we maybe use Angular's `preserverWhiteSpaces` on the component decorator?

DanielPalafox: I applied another solution

leomendoza123 (approved):

PR #925, opened 2021-06-21
fix label gap in names

DanielPalafox:


DanielPalafox:
[[QA][public page] the label and content of the name panel should keep their initial relationship as the panel grows](https://trello.com/c/u6n65vM5/7380-qapublic-page-the-label-and-content-of-the-name-panel-should-keep-their-initial-relationship-as-the-panel-grows)

leomendoza123 (approved):

PR #950, opened 2021-07-22
fix org name and resource type

DanielPalafox:


DanielPalafox:
[[QA][public page] Org name in all caps changes to capitalized case in research resources](https://trello.com/c/wrynkieI/7429-qapublic-page-org-name-in-all-caps-changes-to-capitalized-case-in-research-resources)

leomendoza123 (approved):

PR #998, opened 2021-08-12
Fix show more button behaviour

leomendoza123:


leomendoza123:
[[QA][public page] Research resource details not shown when switching between sources](https://trello.com/c/miaelstR/7497-qapublic-page-research-resource-details-not-shown-when-switching-between-sources)

PR #1000, opened 2021-08-17
add missing make a copy edit functionality

DanielPalafox:


DanielPalafox:
[[speedup Q3] Affiliations: Edit your own version (2)](https://trello.com/c/6fzj6etO/7389-speedup-q3-affiliations-edit-your-own-version-2)

src/app/cdk/panel/panel/panel.component.ts
```diff
@@ -117,7 +120,7 @@ export class PanelComponent implements OnInit {
     return obj && (obj.value || obj.putCode)
   }
 
-  openModal() {
+  openModal(isCopy: boolean) {
```
leomendoza123, line 123:
```suggestion
  openModal(options:{ createACopy: boolean = false} ) {
```
This makes this option easier to understand when reading the code from the template. Also might only be required when the user is making a copy by defaulting to false.

src/app/record/components/work-stack/work-stack.component.ts
```diff
@@ -134,5 +137,12 @@ export class WorkStackComponent implements OnInit {
     return item.putCode.value
   }
 
+  userIsSource(work: Work): boolean {
+    if (this.userInfo) {
+      return work.source === this.userInfo.REAL_USER_ORCID;
```
leomendoza123: I believe `REAL_USER_ORCID` represents the logged-in user, so this will be a problem with delegate users.  I think you need to use `EFFECTIVE_USER_ORCID`, please test this first.

leomendoza123: Check this on affiliations and fundings as well

leomendoza123 (changes requested):

src/app/record/components/work-stack/work-stack.component.ts
DanielPalafox: You are right!. I will fix it

src/app/cdk/panel/panel/panel.component.ts
DanielPalafox: makes sense. I will change it

leomendoza123 (approved):

PR #1016, opened 2021-08-25
fix modal content

DanielPalafox:


DanielPalafox:
it is already fixed in another PR

PR #1023, opened 2021-08-30
fix user OBO not been shown

leomendoza123:


leomendoza123:
[OBO source is missing from the new public page.](https://trello.com/c/R0Z2JvR7/7525-obo-source-is-missing-from-the-new-public-page)

DanielPalafox (approved):

PR #1148, opened 2021-09-20
Fix cypress tests

wjrsimpson:


leomendoza123 (commented):
@wjrsimpson do we want to remove some of the other `.only` functions we still have as part of this PR?

wjrsimpson:
@leomendoza123 Ah, good thinking. How does it look now?

leomendoza123 (approved):

PR #1150, opened 2021-09-20
Modify padding

amontenegro:


src/app/cdk/panel/panels/panels.component.scss
```diff
@@ -9,8 +9,8 @@
   justify-content: space-between;
   align-items: center;
   min-height: 56px;
-  padding-left: 14px;
-  padding-right: 14px;
+  padding-left: 8px;
```
leomendoza123, line 12:
```suggestion
 padding-inline-start: 8px;
```

src/app/cdk/panel/panels/panels.component.scss
```diff
@@ -9,8 +9,8 @@
...
+  padding-right: 16px;
```
leomendoza123, line 13:
```suggestion
 padding-inline-end: 16px;
```

leomendoza123 (changes requested):
these requested changes will save us problems with RTL languages

leomendoza123 (approved):

PR #1154, opened 2021-09-20
Set default visibility on new elements

amontenegro:


amontenegro:
[[QA][my-orcid] When manually adding activities, the visibility selector should be set to the user's default visibility](https://trello.com/c/NqaU5gbq/7559-qamy-orcid-when-manually-adding-activities-the-visibility-selector-should-be-set-to-the-users-default-visibility)

leomendoza123 (approved):

PR #1159, opened 2021-09-21
Hide locked record

leomendoza123:


leomendoza123:
[Locked and deactivated records should not display any information on the public record page.](https://trello.com/c/eoWHCenE/7521-locked-and-deactivated-records-should-not-display-any-information-on-the-public-record-page)

DanielPalafox (approved):

PR #1176, opened 2021-09-29
fix editing employment and qualifications activities

DanielPalafox:


DanielPalafox:
[[QA][my-orcid] Editing a distinction turns it into an invited position](https://trello.com/c/udOlt6R5/7619-qamy-orcid-editing-a-distinction-turns-it-into-an-invited-position)

leomendoza123 (approved):

PR #1183, opened 2021-09-29
fix: github action is running in every PR

DanielPalafox:


leomendoza123:
@DanielPalafox We do not want Storybook running since we are not maintaining on consuming it

DanielPalafox:
Oh, ok. @ leomendoza123 I am gonna disable the action to stop receiving unnecessary notifications about the storybook

PR #1194, opened 2021-10-06
Funding fixes

Camelia-Orcid:


Camelia-Orcid:
[Modal funding broken because of endDateValidator](https://trello.com/c/kGQVrtED/7654-modal-funding-broken-because-of-enddatevalidator)

Camelia-Orcid:
[[QA][my-orcid] Deleted funding only disappears after reloading the page](https://trello.com/c/ALPxEKXz/7640-qamy-orcid-deleted-funding-only-disappears-after-reloading-the-page)

src/app/shared/validators/date/date.validator.ts
```diff
@@ -73,6 +73,28 @@ export function dateMonthYearValidator(dateType: string) {
...
 
+export function endDateMonthYearValidator() {
+  return (c: AbstractControl): { [key: string]: any } | null => {
+    const endDateYear = c.get('endDateGroup.endDateYear').value
+    const endDateMonth = c.get('endDateGroup.endDateMonth').value
+
+    const startDateYear = c.get('startDateGroup.startDateYear').value
+    const startDateMonth = c.get('startDateGroup.startDateMonth').value
+
+    if (!endDateYear && !startDateYear) {
```
leomendoza123, line 84: It seems like the users on the current UI can set only the start or the end date, it seems to me that this code assumes both days are always set. 

```suggestion
    if (!endDateYear || !startDateYear) {
```

Maybe I am reading something wrong? Please verify

leomendoza123 (changes requested):
Please double check this

PR #1198, opened 2021-10-06
fix: url wrapping and update url regex to accept `%` in the urls

DanielPalafox:
see trello card for details

fixes issue 7582


leomendoza123 (changes requested):
Just a small improvement

PR #1230, opened 2021-10-13
Fix duplicate-email-not-detected-when-entered-with-a-different-letter-case

cryptalith:


DanielPalafox:
[[QA][my-orcid] Fix position of error messages in the emails modal](https://trello.com/c/RHuFrIoG/7679-qamy-orcid-fix-position-of-error-messages-in-the-emails-modal)

PR #1241, opened 2021-10-18
diferent duration to get the user status when tab active vs inactive

Camelia-Orcid:


Camelia-Orcid:
[Decrease the userStatus.json calls if the tab is not active (2)](https://trello.com/c/spx2eKg9/7516-decrease-the-userstatusjson-calls-if-the-tab-is-not-active-2)

PR #1245, opened 2021-10-18
fix: Set default sort asc to false in research resources and works

DanielPalafox:
see trello card for details

fixes issue 7688


DanielPalafox:
[[QA][my-orcid] Fix default sorting of research resources](https://trello.com/c/07KlfTdY/7688-qamy-orcid-fix-default-sorting-of-research-resources)

PR #1255, opened 2021-10-20
Ignore white spaces and letter case for the grouping

cryptalith:


leomendoza123:
[[QA] Equal names are not grouped on the new public page [1]](https://trello.com/c/IYcdQBV6/7548-qa-equal-names-are-not-grouped-on-the-new-public-page-1)

PR #1306, opened 2021-11-02
Fixed the missing language name for works

Camelia-Orcid:


Camelia-Orcid:
@leomendoza123  the only way I could make it appear was getting rid of h3 and span. If there is a more elegant solution that you can suggest I would be more than happy to adopt it :)

Camelia-Orcid:
[[QA][my-orcid] Language of translated title not displayed in work details](https://trello.com/c/kLSJtP04/7715-qamy-orcid-language-of-translated-title-not-displayed-in-work-details)

leomendoza123 (approved):

PR #1311, opened 2021-11-02
Fix/blocker (part-of relation) validation on works modal 

leomendoza123:


DanielPalafox (approved):

leomendoza123:
[[QA][my-orcid] No error message when trying to create work with version-of identifier but no self identifier](https://trello.com/c/Wx5HvP7F/7748-qamy-orcid-no-error-message-when-trying-to-create-work-with-version-of-identifier-but-no-self-identifier)

PR #1325, opened 2021-11-04
Attempt to fix the bulk delete error that cannot be reproduced in dev

Camelia-Orcid:


Camelia-Orcid:
[[QA][my-orcid] Error printed in browser console when bulk deleting works](https://trello.com/c/WK0gfoBG/7765-qamy-orcid-error-printed-in-browser-console-when-bulk-deleting-works)

leomendoza123 (approved):

PR #1328, opened 2021-11-04
Fix validations on pubmed

leomendoza123:


amontenegro (approved):

leomendoza123:
[[QA][my-orcid] Unable to add work from PMID or PMCID identifier value](https://trello.com/c/5VNLnplp/7770-qamy-orcid-unable-to-add-work-from-pmid-or-pmcid-identifier-value)

PR #1329, opened 2021-11-05
Fix/7704 qamy orcid new functionality for contributors in works not implemented

DanielPalafox:


DanielPalafox:
[[QA][my-orcid] New functionality for contributors in works not implemented](https://trello.com/c/t8fa5jir/7704-qamy-orcid-new-functionality-for-contributors-in-works-not-implemented)

src/app/record/components/work/work.component.html
```diff
@@ -15,6 +15,28 @@
     <app-display-external-ids
       [externalIdentifiers]="work.workExternalIdentifiers"
     ></app-display-external-ids>
+    <div class="general-data" *ngIf="work?.contributors?.length > 0">
+      <span i18n="#upperCase@@works.contributors"> Contributors </span>:
```
leomendoza123, line 19: the `upperCase` will make `Contributors` be displayed as  `CONTRIBUTORS` on QA, is this what we want?

leomendoza123: We can remove this to avoid it been show in UPPER_CASE

leomendoza123 (changes requested):
Just a small change

src/app/record/components/work/work.component.html
leomendoza123: Sorry Dani! Looking at the figma file 
image
We DO need this to be upper case

leomendoza123 (approved):

PR #1340, opened 2021-11-08
fix: update href attribute to the correct article

DanielPalafox:


DanielPalafox:
[[QA][my-orcid] Link inside export works modal goes to wrong support article](https://trello.com/c/aFblAxBs/7781-qamy-orcid-link-inside-export-works-modal-goes-to-wrong-support-article)

amontenegro (approved):

PR #1348, opened 2021-11-09
Clean cache/remove zoombie sessions

leomendoza123:


leomendoza123:
[[QA][my-orcid] Private user info retained after signing out and visible to user that signs in next](https://trello.com/c/QuwBD39Z/7783-qamy-orcid-private-user-info-retained-after-signing-out-and-visible-to-user-that-signs-in-next)

DanielPalafox (approved):

PR #1369, opened 2021-11-11
Fixed the menu item truncated text

Camelia-Orcid:


Camelia-Orcid:
[[QA][my-orcid] Translated text is truncated in fixed-width dropdowns](https://trello.com/c/3ulKWT8U/7802-qamy-orcid-translated-text-is-truncated-in-fixed-width-dropdowns)

PR #1377, opened 2021-11-11
Fix/314 qamy orcid duplicate email not detected when entered with a different letter case

DanielPalafox:


DanielPalafox:
[[QA][my-orcid] Duplicate email not detected when entered with a different letter case](https://trello.com/c/Y8jjsoxO/7593-qamy-orcid-duplicate-email-not-detected-when-entered-with-a-different-letter-case)

src/app/core/record-emails/record-emails.service.ts
```diff
@@ -49,8 +49,8 @@ export class RecordEmailsService {
     } else if (!options.forceReload) {
       return this.$emailsSubject
     }
-    if (options.cleanCacheIfExist && this.$emailsSubject) {
-      this.$emailsSubject.next(<EmailsEndpoint>undefined)
+    if (options.cleanCacheIfExist) {
```
leomendoza123, line 52: @DanielPalafox please check this, to call `next` on the `this.$emailsSubject` it must not be null. So this should not make a big difference although checking if the `this.$emailsSubject` is not really require since is created on line 48

leomendoza123 (approved):

PR #1387, opened 2021-11-17
Fixed the wrong route for self service

Camelia-Orcid:


Camelia-Orcid:
[[QA] Redirect problem when user clicks member tools in the account menu](https://trello.com/c/DyWn5CZS/7819-qa-redirect-problem-when-user-clicks-member-tools-in-the-account-menu)

amontenegro (approved):

PR #1395, opened 2021-11-25
Moved the display modal logic inside subscribe ( not pipe)

Camelia-Orcid:
@DanielPalafox  can you please review. I tested and works but maybe there were some reasons why it was on pipe and not inside subscribe that I am not aware of. Thanks.


Camelia-Orcid:
[Verify email popover is not displayed sometimes](https://trello.com/c/XbnyR3mb/7826-verify-email-popover-is-not-displayed-sometimes)

DanielPalafox (approved):

PR #1398, opened 2021-11-25
check for undefined instead of empty to emit event when last review is deleted and the array is empty

Camelia-Orcid:
…


Camelia-Orcid:
[[QA][my-orcid] When I delete the only peer review that I have it's still displayed in my record until I reload the page](https://trello.com/c/K3fFkty1/7832-qamy-orcid-when-i-delete-the-only-peer-review-that-i-have-its-still-displayed-in-my-record-until-i-reload-the-page)

amontenegro (approved):

PR #1399, opened 2021-11-25
Fix NPE when website.urlName is null

amontenegro:


amontenegro:
[[QA][my-orcid] Don't allow entering URLs without a name in the websites section](https://trello.com/c/zRUn9y9g/7835-qamy-orcid-dont-allow-entering-urls-without-a-name-in-the-websites-section)

PR #1404, opened 2021-12-02
fix: Keep pagination if a work visibility is changed

DanielPalafox:


DanielPalafox:
[[QA][my-orcid] When I change the visibility setting on a work while paginating, the list reverts to the first page](https://trello.com/c/aAYR7U6M/7837-qamy-orcid-when-i-change-the-visibility-setting-on-a-work-while-paginating-the-list-reverts-to-the-first-page)

leomendoza123 (approved):

PR #1442, opened 2022-01-19
Feature/account settings email frequency

leomendoza123:
Requires https://github.com/ORCID/orcid-ansible_inventory-private/pull/89


src/app/account-settings/components/settings-defaults-email-frequency/settings-defaults-email-frequency.component.html
```diff
@@ -0,0 +1,120 @@
...
+    <label class="row">
+      <strong i18n="@@account.itemsAddedOrEdited"
+        >Items added or edited in your record by a trusted party</strong
+      >
+    </label>
+    <mat-form-field
+      appearance="outline"
+      class="mat-form-field-min input-container"
+    >
+      <mat-select formControlName="send_change_notifications">
```
DanielPalafox, line 42: @leomendoza123 could you rename all the variables using camel case convention

src/app/core/user/user.service.ts
```diff
@@ -78,7 +78,7 @@ export class UserService {
   private ONE_MINUTE = 60 * 1000
   private FIVE_MINUTES = 5 * 60 * 1000
   private interval$: BehaviorSubject<number> = new BehaviorSubject<number>(
-    this.ONE_MINUTE
+    10000
```
DanielPalafox, line 81: @leomendoza123 Could you create a variable as it was before so it'd be more readable

DanielPalafox (changes requested):

PR #1447, opened 2022-01-20
Add TwoFactorSigninGuard

auumgn:


src/app/guards/two-factor-signin.guard.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { TestBed } from '@angular/core/testing'
+import { TwoFactorSigninGuard } from './two-factor-signin.guard'
+
+describe('LinkAccountGuard', () => {
```
leomendoza123, line 4:
```suggestion
describe('TwoFactorSigninGuard', () => {
```

auumgn:
[social sign in doesn't work if 2fa is enabled on the account](https://trello.com/c/RwDKqebn/7201-social-sign-in-doesnt-work-if-2fa-is-enabled-on-the-account)

leomendoza123 (approved):

PR #1449, opened 2022-01-20
fix: Display contributor roles in public page

DanielPalafox:


src/app/record/components/work/work.component.ts
```diff
@@ -100,21 +100,25 @@ export class WorkComponent implements OnInit {
     contributors.forEach((c) => {
       if (this.isPublicRecord) {
         if (c?.orcid?.value === this.isPublicRecord) {
```
leomendoza123, line 102: maybe we can simplify this 3 `ifs` into a single

`c?.orcid?.value === this.isPublicRecord || this.id`

amontenegro (approved):

PR #1456, opened 2022-01-26
Fixed the ISNI URL

Camelia-Orcid:
The ISNI url has been fixed and the other identifiers that we don't have yet a base URL for are displayed as IDs without a link.


Camelia-Orcid:
[ISNI org identifiers url is incorrect.](https://trello.com/c/dpjTdzYK/7898-isni-org-identifiers-url-is-incorrect)

leomendoza123 (approved):

PR #1463, opened 2022-01-27
Make config.json calls happen at the same time that userStatus calls

leomendoza123:


leomendoza123:
[Lets call user status checks less frequently](https://trello.com/c/w44T1q7R/7902-lets-call-user-status-checks-less-frequently)

PR #1470, opened 2022-02-01
Fix togglz update not working after headless signin

leomendoza123:


src/app/core/user/user.service.ts
```diff
@@ -64,7 +64,7 @@ export class UserService {
     private _disco: DiscoService,
     private _userInfo: UserInfoService
   ) {}
-  $userStatusChecked = new Subject()
```
leomendoza123, line 67: When the user arrives to a headless login (AKA Oauth url page) nothing there will ask for any togglz state.  This means that the `$userStatusChecked` is triggered, but the togglz is not listening to it to call the `config.json` endpoint. 

With this change, even if the `getTogglz()` gets called later on the application life, the call to the config.json will be made.

amontenegro (approved):

auumgn:
[Prevent locked records from participating in oauth](https://trello.com/c/Ga43Bbw2/25-prevent-locked-records-from-participating-in-oauth)

PR #1477, opened 2022-02-03
fix: Authorization screen not skipped if there was an active token

DanielPalafox:


DanielPalafox:
[[QA] Authorization screen not skipped when there is already an active token for the user](https://trello.com/c/358LkBir/7907-qa-authorization-screen-not-skipped-when-there-is-already-an-active-token-for-the-user)

amontenegro (approved):

PR #1481, opened 2022-02-03
Fix/7915 qa contributor names in works not displayed correctly

DanielPalafox:


DanielPalafox:
[[QA] Contributor roles in summary view only displayed in public page if record holder is signed in](https://trello.com/c/iD62b616/7909-qa-contributor-roles-in-summary-view-only-displayed-in-public-page-if-record-holder-is-signed-in)

PR #1490, opened 2022-02-09
Added additional fix for preferred identifiers and base url for wikidata

Camelia-Orcid:


Camelia-Orcid:
[ISNI org identifiers url is incorrect.](https://trello.com/c/dpjTdzYK/7898-isni-org-identifiers-url-is-incorrect)

src/app/record/components/org-identifier/org-identifier.component.html
```diff
@@ -66,9 +66,34 @@ <h3 class="orc-font-body-small">
...
+            target="_blank"
+            rel="noopener noreferrer"
+            href="{{ getLink(org.identifierType, org.preferred) }}"
+          >
+            {{ org.preferred }}
+          </a>
+        </ng-container>
+        <ng-template #plainUrl>
+          <a
+            *ngIf="isUrl(org.preferred)"
```
leomendoza123, line 87: seems like `*ngIf="isUrl(org.preferred)"` will never be true, since is wrap with another  if `!isUrl(org.preferred)`

leomendoza123 (changes requested):
Just a comment to check

leomendoza123 (approved):

PR #1501, opened 2022-02-11
7940 qa change column name in alternate sign in accounts section

leomendoza123:


leomendoza123:
[[QA] Change column name in alternate sign in accounts section](https://trello.com/c/0tqEN58p/7940-qa-change-column-name-in-alternate-sign-in-accounts-section)

DanielPalafox (approved):

PR #1506, opened 2022-02-15
Feature/7927 account actions remove a duplicate record 5

leomendoza123:


src/app/two-factor/two-factor.module.ts
```diff
@@ -1,29 +1,30 @@
...
+import { TwoFactorAuthenticationFormModule } from '../cdk/two-factor-authentication-form/two-factor-authentication-form.module'
+import { TwoFactorComponent } from './pages/two-factor/two-factor.component'
+import { TwoFactorRoutingModule } from './two-factor-routing.module'
 
 @NgModule({
   declarations: [TwoFactorComponent],
   imports: [
     CommonModule,
     TwoFactorRoutingModule,
-    SignInModule,
```
leomendoza123, line 19: I did move the `old` two-factor component into its own module `TwoFactorAuthenticationFormModule` so we don't have to import the huge `SignInModule` to consume it.

src/app/account-settings/components/dialog-actions-duplicated-two-factor-auth/dialog-actions-duplicated-two-factor-auth.component.ts
```diff
@@ -0,0 +1,49 @@
...
+        orcid: this.data.deprecatingOrcid,
+        verificationCode: code.verificationCode || null,
+        recoveryCode: code.recoveryCode || null,
+        errors: [],
+        redirectUrl: null,
+      })
+      .subscribe((res) => {
+        this.loading = false
+        if (res.errors && res.errors.length > 0) {
+          if (res.errors[0].includes('verification')) {
```
leomendoza123, line 39: We should not use the `res.errors[0].includes('verification')` to check the 2FA response errors types since this will only work in English.
This is an issue we need to solve on the backend, by sending back an error code and not a string, so I will create a separate card to solve this on all the 2FA places.

leomendoza123:
[Account actions: Remove a duplicate record (5)](https://trello.com/c/8WTLXV51/7927-account-actions-remove-a-duplicate-record-5)

src/app/account-settings/components/dialog-actions-duplicated/dialog-actions-duplicated.component.html
```diff
@@ -0,0 +1,104 @@
+<app-modal [noSidebar]="true">
+  <app-modal-header i18n="@@account.removeDuplicateRecod"
```
DanielPalafox, line 2: Could you fix the typo here in the property

DanielPalafox (changes requested):

DanielPalafox (approved):

PR #1514, opened 2022-02-21
feature: Add Appcues to only qa and sbox

DanielPalafox:


src/app/record/pages/my-orcid/my-orcid.component.ts
```diff
@@ -124,6 +138,9 @@ export class MyOrcidComponent implements OnInit, OnDestroy {
             this._robotsMeta.disallowRobots()
           }
           this._openGraph.addOpenGraphData(userRecord, { force: true })
+          if (environment.APPCUES) {
+            this.initializeAppCues(this.userInfo, this.userRecord)
```
leomendoza123, line 142: Should this be wrapped with an `if`, that only calls the `initializeAppCues` once? 

At the moment this will be called every time an endpoint responds, which results in calling the app cues `identity` function multiple times with potentially different values. 

What do you think? 🤔

leomendoza123 (changes requested):
Requesting a change to avoid calling the identity function multiple times

src/app/record/pages/my-orcid/my-orcid.component.ts
DanielPalafox: done!

leomendoza123 (approved):

PR #1519, opened 2022-02-22
fix: Display error messages in works to import using bibtex

DanielPalafox:


DanielPalafox:
[[QA] Works imported from BibTeX containing invalid URLs are not saved but there's no error message](https://trello.com/c/94UhxS14/7957-qa-works-imported-from-bibtex-containing-invalid-urls-are-not-saved-but-theres-no-error-message)

src/app/core/record-works/record-works.service.ts
```diff
@@ -206,7 +206,7 @@ export class RecordWorksService {
       )
   }
 
-  save(work: Work, bibtex?: boolean) {
+  save(work: Work, bibtex?: boolean): Observable<any> {
```
leomendoza123, line 209: Please define a response object. 
Maybe a simple generic error object is ok like
```suggestion
  save(work: Work, bibtex?: boolean): Observable<{ errors :string}> {
```

src/app/record/components/work-stack-group/modals/work-bibtex-modal/work-bibtex-modal.component.ts
```diff
@@ -85,6 +86,10 @@ export class WorkBibtexModalComponent implements OnInit, OnDestroy {
                   .subscribe((data) => {
                     data.forEach((work) => {
                       that.worksFromBibtex.push(work)
+                      if (work.errors.length > 0 && !that.isAnInvalidWork) {
+                        that.isAnInvalidWork = true
```
leomendoza123, line 90: Wondering what happens if the user enters an invalid BibTeX and after that, the user fixes it with a valid BibTeX? 

It seems like we never set the `isAnInvalidWork` back to false 🤔

leomendoza123 (changes requested):
Please check my 2 comments

src/app/record/components/work-stack-group/modals/work-bibtex-modal/work-bibtex-modal.component.ts
DanielPalafox: Every time you want to import a new BibTeX file you need to click cancel and open the modal again

leomendoza123 (approved):

leomendoza123:
[[QA] Works imported from BibTeX containing invalid URLs are not saved but there's no error message](https://trello.com/c/94UhxS14/7957-qa-works-imported-from-bibtex-containing-invalid-urls-are-not-saved-but-theres-no-error-message)

PR #1521, opened 2022-02-23
fix: Display backend error messages inside old password mat form field

DanielPalafox:


DanielPalafox:
[[QA] Change position of error when entering incorrect password in the account password section](https://trello.com/c/kblmClxP/7964-qa-change-position-of-error-when-entering-incorrect-password-in-the-account-password-section)

src/app/account-settings/components/settings-security-password/settings-security-password.component.ts
```diff
@@ -70,12 +66,12 @@ export class SettingsSecurityPasswordComponent implements OnInit, OnDestroy {
           this.loading.emit(false)
 
           if (!value.password) {
-            this.errors = []
             this.form.reset()
             this.success = true
           } else {
-            this.success = false
-            this.errors = value.errors
+            this.form.controls['oldPassword'].setErrors({
```
leomendoza123, line 72: Is this error removed when the user updated the value of the password? 
Just asking to make sure this error is remove at some point

leomendoza123 (changes requested):
Please check my comment maybe not change is required

leomendoza123 (approved):

PR #1535, opened 2022-03-02
fix: Remove case insensitive in ORCID REGEXP

DanielPalafox:


Camelia-Orcid:
[Redirect records with lowercase checksum digit to the right X url](https://trello.com/c/HdI0rPfV/7977-redirect-records-with-lowercase-checksum-digit-to-the-right-x-url)

amontenegro (approved):

PR #1538, opened 2022-03-07
fix: Remove unnecessary force load option in public-record.json endpoint

DanielPalafox:


leomendoza123 (changes requested):
This breaks more than what it fixes... 
I will open a separate PR for this tomorrow. 

We can lose this PR

PR #1539, opened 2022-03-09
Sleep force reloads for public page endpoint 

leomendoza123:


leomendoza123:
[[DEVS] Calling too many time the `public-record.json` endpoint](https://trello.com/c/6o272lHI/7992-devs-calling-too-many-time-the-public-recordjson-endpoint)

PR #1552, opened 2022-03-18
7842 rors are included in search for the UI and the api 

Camelia-Orcid:
ROR ids start with http, we assume that if starts with http is a ror id and map it to ror-org-id in the institution name


Camelia-Orcid:
[ROR's are included in search for the UI and the API](https://trello.com/c/4gta7Cwq/7842-rors-are-included-in-search-for-the-ui-and-the-api)

amontenegro (approved):

PR #1571, opened 2022-04-05
Fix/8050 long org names truncated by fixed width typeahead

leomendoza123:


leomendoza123:
[Long org names truncated by fixed width typeahead](https://trello.com/c/XQbhjUSO/8050-long-org-names-truncated-by-fixed-width-typeahead)

PR #1582, opened 2022-04-07
fix: Do not refresh sign in page if forceLogin parameter is present

DanielPalafox:


DanielPalafox:
[[QA] Signed in user caught in endless redirect when visiting OAuth link with openid + prompt=login](https://trello.com/c/8YADz7vW/8053-qa-signed-in-user-caught-in-endless-redirect-when-visiting-oauth-link-with-openid-promptlogin)

amontenegro (approved):

PR #1596, opened 2022-04-21
Added the hide reveal citations

Camelia-Orcid:
because citation-js is not a ESModule I had troubles resolving the dependencies properly. I reused the same library we already have in works to parse bibtex ( bibtex-parse-js ) and a pre tag.


Camelia-Orcid:
[Citation hide & reveal](https://trello.com/c/C2y57qxk/7974-citation-hide-reveal)

leomendoza123 (approved):

PR #1603, opened 2022-04-26
Fixed the QA errors as detailed in the card

Camelia-Orcid:
in the sub-header "Citation (citation-type)", the citation type should not be in bold. Here's an example of how it should be displayed: "Citation (bibtex)"

The link "switch to expanded formatting" comes right after the end of the citation. There should be a line break between the end of the citation and the link. As seen here: capture2-citation.jpg

When a user clicks the link "switch to expanded formatting", the text of the citation should wrap into the next line. Right not it overflows the screen. As seen in: capture1-citation.jpg

### DIFF
--- a/src/app/record/components/work/work.component.html
+++ b/src/app/record/components/work/work.component.html
@@ -317,6 +317,7 @@
           </ng-container>
           <ng-container *ngIf="work.citation?.citationType?.value === 'bibtex'">
             <br />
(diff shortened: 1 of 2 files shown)


Camelia-Orcid:
[Citation hide & reveal](https://trello.com/c/C2y57qxk/7974-citation-hide-reveal)

amontenegro (approved):

PR #1612, opened 2022-05-10
fix: Add placeholder Names is private in contributor details list

DanielPalafox:


DanielPalafox:
[Placeholder for private names in contributor lists](https://trello.com/c/rQnsZQbc/8094-placeholder-for-private-names-in-contributor-lists)

src/app/record/components/work/work.component.html
```diff
@@ -33,15 +33,18 @@
...
+            contributor?.creditName?.content !== privateName
           "
         >
           {{ contributor?.creditName?.content + (!last ? '; ' : '') }}
         </ng-container>
         <ng-container
           *ngIf="
             (!contributor?.creditName ||
-              contributor?.creditName?.content === '') &&
+              contributor?.creditName?.content === '' ||
```
leomendoza123, line 46: Looking at the backend logic 
https://github.com/ORCID/ORCID-Source/pull/6533/files#diff-95458e03beba5286beb751702a1fd06f126e97d1f1c4b52bf4a30d7d9c530d73R205

We don't need to check `contributor?.creditName?.content === ''` since it will never be true

leomendoza123: (also for line 38)

src/app/record/components/work/work.component.html
```diff
@@ -183,7 +186,7 @@ <h3 class="orc-font-body-small" i18n="@@works.contributors">
       >
         <ng-container *ngIf="index < maxNumberContributorsWorkDetails">
           <ng-container *ngIf="contributor?.creditName?.content">{{
-            contributor?.creditName?.content
+            isNamePrivate(contributor.creditName.content)
```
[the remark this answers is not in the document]
leomendoza123: The better way to do it is to map the data as it comes back from the service on the `ngOnInit`

leomendoza123 (changes requested):

src/app/record/components/work/work.component.html
DanielPalafox: @leomendoza123 Great catch. I will remove the unnecessary validation

src/app/record/components/work/work.component.html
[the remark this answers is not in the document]
DanielPalafox: @leomendoza123 Ohh my bad. I will fix it so it doesn't do unnecessary calls

leomendoza123 (approved):

PR #1623, opened 2022-05-16
Remove success message from reset password

leomendoza123:


src/app/reset-password/reset-password/reset-password.component.html
```diff
@@ -197,14 +197,6 @@
                 </mat-error>
               </mat-form-field>
 
-              <div
-                *ngIf="success === true"
-                class="error-container"
-                i18n="@@account.yourPasswordHasBeenUpdate"
```
DanielPalafox, line 203: @leomendoza123 Could you please also remove the unused property from the `*.properties` files

DanielPalafox (changes requested):

DanielPalafox (approved):

PR #1649, opened 2022-06-06
fix 8162

leomendoza123:


leomendoza123:
[[QA] Duplicate URL error message only shown for one of the duplicates](https://trello.com/c/Wl3iARyK/8162-qa-duplicate-url-error-message-only-shown-for-one-of-the-duplicates)

PR #1709, opened 2022-08-03
fix(work): Update unique validator to allow duplicated options

DanielPalafox:
[`No specified role` should not trigger a 'multiple Credit role' error](https://trello.com/c/G12MG7m6/8239-no-specified-role-should-not-trigger-a-multiple-credit-role-error)


DanielPalafox:
[`No specified role` should not trigger a 'multiple Credit role' error](https://trello.com/c/G12MG7m6/8239-no-specified-role-should-not-trigger-a-multiple-credit-role-error)

amontenegro (approved):

PR #1716, opened 2022-08-08
oauth tc 34 missing client id

kprisAutomation:


cypress/e2e/oauth/oauth_tc34.cy.js
```diff
@@ -0,0 +1,31 @@
...
+pre-condition: authorization link doesn't contain client_id parameter
+1 - visit authorization link
+expected: user is taken to OAuth error page "oauth_error / Please specify a client id"
+*/
+
+describe('OAuth cypress tests', async function () {
+  const recordOwner = userData.cyOAuth_RecordOwnerTC_Sessions
+  const scope = '/person/update'
+  const authorizationLink =
+    'https://qa.orcid.org/oauth/authorize?client_id=' +
```
amontenegro, line 16: @Karen-Automation what happen if the `client_id` param is not included at all? are we testing that in other TC?

amontenegro (changes requested):

cypress/e2e/oauth/oauth_tc34.cy.js
kprisAutomation: you are totally right, the param should not be preset at all. Will work on the fix.

kprisAutomation: Done, now we validate both scenarios: empty and missing client id

amontenegro (approved):

PR #1725, opened 2022-08-11
fix: Rollback display original role if not translation is founded

DanielPalafox:


DanielPalafox:
[[QA] Non-credit roles are not displayed when copy and editing work added via API](https://trello.com/c/kbwCGPen/8250-qa-non-credit-roles-are-not-displayed-when-copy-and-editing-work-added-via-api)

amontenegro (approved):

PR #1726, opened 2022-08-11
tc#44 oauth link contains language param

kprisAutomation:


amontenegro:
@kprisAutomation shouldnt we check that the user is not logged in before running the TC?

amontenegro (changes requested):
@kprisAutomation shouldnt we check that the user is not logged in before running the TC?

amontenegro (approved):

PR #1740, opened 2022-09-07
Manage legacy role translations for role pipes

leomendoza123:


src/app/record/components/work-contributor-role/work-contributor-roles.component.html
```diff
@@ -14,11 +14,7 @@
       <div class="row role">
         <div class="col l9 m8 s4 no-gutters orc-font-body-small">
           <ng-container *ngIf="role.disabled">
-            {{
-              role.get('role').value
-                | roleTranslation: contributionRoles
-                | titlecase
-            }}
+            {{ role.get('role').value | roleTranslation | titlecase }}
```
amontenegro, line 17: @leomendoza123 could this `role.get('role')` be ever be null? if so, this line could throw an NPE right?

amontenegro (changes requested):

src/app/record/components/work-contributor-role/work-contributor-roles.component.html
leomendoza123: This should never be null. 
I can be null only on a bizarre scenario where on the call to GET `/works/getWorkInfo.json?workId=<WORK_ID>` gets back a `contributorsGroupedByOrcid` like the following. 

```
  "contributorsGroupedByOrcid": [
    {
      "contributorOrcid": {
        "uri": "https://qa.orcid.org/0000-0002-9361-1905",
        "path": "0000-0002-9361-1905",
        "host": null
      },
      "creditName": {
        "content": "Leonardo"
      },
      "contributorEmail": null,
      "contributorAttributes": null,
      "rolesAndSequences": [
        {
          "contributorSequence": null,
          "contributorRole": "conceptualization" // <----- THIS WOULD HAVE TO BE NULL ON THE BACK
        }
      ]
    }
  ],
  ```
  
  This hypothetical scenario would be a backend issue where the role objects are not correctly returned.

amontenegro (approved):

amontenegro:
https://trello.com/c/n7tH63iW/8266-qa-work-displays-an-undefined-role

PR #1746, opened 2022-09-15
fix: Update contributor name max length validation

DanielPalafox:


DanielPalafox:
[[QA] Other contributor name field should be less than 100chars](https://trello.com/c/c5Dt0vvu/8289-qa-other-contributor-name-field-should-be-less-than-100chars)

amontenegro (approved):

PR #1752, opened 2022-09-20
oauth test case #55 user refreshes access token

kprisAutomation:


cypress/e2e/oauth/oauth_tc55.cy.js
```diff
@@ -0,0 +1,111 @@
...
+      newAccessToken = responseStr.slice(posStart, posEnd)
+      cy.log('new access Token: ' + newAccessToken)
+
+      //Try to use new access token
+      const curlPostWork =
+        "curl -i -H 'Content-type: application/vnd.orcid+xml' -H " +
+        "'Authorization: Bearer " +
+        newAccessToken +
+        "' -d " +
+        "'@/Users/karenmadrigal/Documents/projects/registry/cypress/fixtures/work_minimal.xml' -X" +
```
amontenegro, line 96: @kprisAutomation I think this path should be changed for something else right?

amontenegro (changes requested):

cypress/e2e/oauth/oauth_tc55.cy.js
kprisAutomation: fixed

amontenegro (approved):

PR #1755, opened 2022-09-21
Added the translation for language form field

Camelia-Orcid:


Camelia-Orcid:
[Update the language for chinese that can be used when adding works](https://trello.com/c/22RstTPi/8299-update-the-language-for-chinese-that-can-be-used-when-adding-works)

amontenegro (approved):

PR #1759, opened 2022-09-27
fix: Only call getWorks once when BibTeX are been imported and Add mi…

DanielPalafox:
…ssing tests


DanielPalafox:
[[QA] `works/worksExtendedPage.json` endpoint is being called every time a work of a BibTeX file is created](https://trello.com/c/vf33IJX8/8298-qa-works-worksextendedpagejson-endpoint-is-being-called-every-time-a-work-of-a-bibtex-file-is-created)

PR #1761, opened 2022-10-03
fix: Only display contributors count if contributors are gonna be edi…

DanielPalafox:
…table


DanielPalafox:
[(P3.1) Add contributors via identifier (DOI & PubMed ID)](https://trello.com/c/MUTV4zba/8305-p31-add-contributors-via-identifier-doi-pubmed-id)

PR #1787, opened 2022-11-09
Update Angular 14

leomendoza123:
## Notes for Devs:

- Please note that Angular 14 uses typed Forms. For now to make the migration easier all`FormBuilder` instances are changed to `UntypedFormBuilder`. **But we should use `FormBuilder` from now on for new features or update current features** in this way we get typed forms 
- I recommend using `yarn` this is what we use on GitHub Actions and on the build machines so that will make your local environments more similar to those. 
- Make sure you delete `node_modules`, before running `yarn install` to get the latest updates

## Notes for QA
- We have many style changes across the app. So we need to check nothing looks the wrong color, position, or size. 
- We are updating 3 major versions of the Material libraries, so be extra careful checking that buttons, dropdown menus, icons, dialogs, and typeahead are ok. 
- Let's check the `language cookie`, and check the app remembers the user language.
- Let's ensure the feature to remember the institution user has selected before still works on `institutional login`. 
- Ensure the `cookie banner`  is correctly shown and hidden 
- Ensure the `testing warning` is correctly shown and hidden on Sandbox
- We can expect a few Cypress issues @Karen-Automation we can look at those together


amontenegro (approved):

leomendoza123:
[Upgate the registry to the latest Angular (Angular 14)](https://trello.com/c/013F52Fk/8360-upgate-the-registry-to-the-latest-angular-angular-14)

PR #1807, opened 2022-12-05
fix: Update Turkey translation in all available languages

DanielPalafox:


DanielPalafox:
[[QA] Turkey is not displayed and Turkmenistan is duplicated in the list of countries](https://trello.com/c/IjQDb7d8/8395-qa-turkey-is-not-displayed-and-turkmenistan-is-duplicated-in-the-list-of-countries)

PR #1815, opened 2022-12-13
added style for input placeholders

Camelia-Orcid:


src/styles.scss
```diff
@@ -103,3 +103,15 @@ html {
   clip: rect(0, 0, 0, 0);
   border: 0;
 }
+
```
leomendoza123, line 106: Cami we have a folder `src/assets/scss/material.orcid.overwrites.scss` to put all material overwrites of the app. 

Let's put these overwrites there!

leomendoza123 (approved):

leomendoza123 (changes requested):
Let's use `src/assets/scss/material.orcid.overwrites.scss` for overwrites

src/styles.scss
Camelia-Orcid: created a new file form-field.css for the placeholders overwrites

leomendoza123 (approved):

PR #1821, opened 2022-12-15
Removed the ::ng-deep was causing the color not to be applied and added the placeholder color for search

Camelia-Orcid:
After moving the code to the overwrite directory, the color was not applied.  This commit contains the fix for it.


Camelia-Orcid:
[Accessible placeholder colour](https://trello.com/c/IDocXR4y/8385-accessible-placeholder-colour)

leomendoza123 (approved):

PR #1824, opened 2022-12-15
Fix/8402 qa also known as panel should not be part of tab order in public pag

leomendoza123:


leomendoza123:
[[QA] "Also known as" panel should not be part of tab order in public page](https://trello.com/c/nVX06EgQ/8402-qa-also-known-as-panel-should-not-be-part-of-tab-order-in-public-page)

amontenegro (approved):

PR #1825, opened 2022-12-15
fix: Initialize help hero only if is not public or record with issues

DanielPalafox:


leomendoza123 (approved):

DanielPalafox:
[HelpHero script should not be loaded for logged out users](https://trello.com/c/2x2FuqXK/8401-helphero-script-should-not-be-loaded-for-logged-out-users)

PR #1830, opened 2022-12-16
fix place holder color

leomendoza123:


leomendoza123:
[[QA] Accessible placeholder color not applied to inputs and textareas](https://trello.com/c/lzQ458qH/8410-qa-accessible-placeholder-color-not-applied-to-inputs-and-textareas)

PR #1833, opened 2022-12-19
fix other name aria label

leomendoza123:


leomendoza123:
[[QA] Add aria-label to "Also known as" panel in public page](https://trello.com/c/nVX06EgQ/8402-qa-add-aria-label-to-also-known-as-panel-in-public-page)

amontenegro (approved):

PR #1840, opened 2023-01-02
custom command to log in with 2FA enabled user.

kprisAutomation:
Reading admin user data from .env file. 
Inbox Notif script record owner adds trusted individual. 
Inbox notif admin adds trusted individual to a record.


amontenegro:
@leomendoza123 please review this one.

Thanks

leomendoza123 (approved):

PR #1892, opened 2023-03-02
Fix/8498 implement ga4 events in qa alongside ua events using GTM

DanielPalafox:


DanielPalafox:
[Create a GTM account for QA and setup container so that it can manage UA tag and GA4 tag](https://trello.com/c/1QCwoxmo/8497-create-a-gtm-account-for-qa-and-setup-container-so-that-it-can-manage-ua-tag-and-ga4-tag)

amontenegro (approved):

PR #1910, opened 2023-03-21
feature: Only split contributors if ` and ` contains whitespaces

DanielPalafox:


DanielPalafox:
[Names containing character sequence "and" not displayed correctly in contributor list after adding BibTeX](https://trello.com/c/SdxOIwpY/8408-names-containing-character-sequence-and-not-displayed-correctly-in-contributor-list-after-adding-bibtex)

amontenegro (approved):

Issue #1926, opened 2023-04-06
Double calls to client's Redirect URI breaks client SSO


cslzchen:
**Describe the bug**

When client attempts to do OAuth, after the `authroize` request is successful, ORCiD server returns a 200 (1), in which a script runs and triggers the redirection to the client's configured Redirect URI (2). However, a second request to the Redirect URI is made about 4-5 seconds later (3). The second request causes the browser to cancel the first request.

The issue is, the first request has already cleared the client server of the one-time OAuth SSO state when it is canceled. It is in the process of a few final redirections between our authentication server (which is also the OAuth client, from ORCiD server's perspective) and our app server to finish authentication into our app. Thus, when the second request comes in, it fails SSO due to the invalid state.

Below, I attached both screenshots of doc-only network history and a full network history, a screenshot of the error page and a short video clip.

![prod-orcid-sso-network-requests-doc-only](https://user-images.githubusercontent.com/3750414/230429509-33a46369-e82c-47c8-9136-2a06dbcbd580.png)

![prod-orcid-sso-requests-all](https://user-images.githubusercontent.com/3750414/230430473-b6ae50f7-160a-45ad-b2af-d5a6a3816c7f.png)

![prod-osf-orcid-sso-error-page](https://user-images.githubusercontent.com/3750414/230429459-a5a1abe4-3f36-4e35-8edf-d2952c125bb0.png)

https://user-images.githubusercontent.com/3750414/230430631-f7c9f7fc-65d6-4f5d-ad34-ea6dd610d306.mov

**To Reproduce**

1. Go to "osf.io": https://osf.io/
2. Click on "Sign in" which will redirect you our authentication server "accounts.osf.io": https://accounts.osf.io/login?service=https://osf.io/login/?next=https%253A%252F%252Fosf.io%252F
3. Click "Sign in via ORCiD"
4. Follow the steps and you will see the error.

Note: If this is your first time login into OSF with ORCiD, you will go through a different process that ask you to enter an email and your full name. You will then receive a confirmation email to confirm account creation with ORCiD. After confirmation, log out of OSF to reproduce the error.

**Expected behavior**
`/authorize` should return a 302 to client's redirect URI; or if it returns 200, the script on that page should make only one call the the client's Redirect URI; or if it needs to make another request after some timeout, currently 4 seconds is way too short.

**Screenshots**

See Description

**Desktop (please complete the following information):**

- OS: [e.g. iOS] macOS 12 and 10
[...]

PR #1927, opened 2023-04-07
Prevent redirecting againt in the middle of redirect (ORCID#1926)

kadet1090:
This should hopefully be temporary fix for issue mentioned in #1926 where some of SSO clients like OSF or our [Bridge of Knowledge](https://mostwiedzy.pl] have problems with users being unable to log in into the application. The issue is caused by the race condition of GTM and GTM timeout (4s) so as temporary solution I propose to simply allow only one call to the `sendUserToRedirectURL` which should fix the problem. 

However proper solution is needed because setting 4s timeout as workaround for users that have addons blocking GTM does not seem like proper solution causing unnecessary delay for some users.

### DIFF
--- a/src/app/guards/authorize.guard.ts
+++ b/src/app/guards/authorize.guard.ts
@@ -23,6 +23,7 @@ import { GoogleTagManagerService } from '../core/google-tag-manager/google-tag-m
 })
 export class AuthorizeGuard implements CanActivateChild {
   lastRedirectUrl: string
+  inTheMiddleOfRedirect: boolean = false
   constructor(
     private _user: UserService,
     private _router: Router,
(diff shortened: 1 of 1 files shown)


src/app/guards/authorize.guard.ts
```diff
@@ -67,7 +68,14 @@ export class AuthorizeGuard implements CanActivateChild {
   }
 
   sendUserToRedirectURL(oauthSession: RequestInfoForm): Observable<boolean> {
+    // Prevent double redirection caused by races using simple semaphore like construct
+    if (this.inTheMiddleOfRedirect) {
+      return;
+    }
+
     this.window.location.href = oauthSession.redirectUrl
+    this.inTheMiddleOfRedirect = true;
```
leomendoza123, line 77: The problem is that we have 2 redirects, one is called by GTM library it self, and the other one is called by `sendUserToRedirectURL`. 

So in reality the `sendUserToRedirectURL` is only call once, and that is why this boolean check does not work .

PR #1933, opened 2023-04-17
Update tr name

amontenegro:


amontenegro:
['Turkey' is now called 'Türkiye'](https://trello.com/c/xEdONy3D/8566-turkey-is-now-called-t%C3%BCrkiye)

PR #1975, opened 2023-06-05
move-back-to-trusted-only-string

cryptalith:


leomendoza123:
[[QA] Fix translation for visibility setting label for "trusted"](https://trello.com/c/DW1U1MPD/8653-qa-fix-translation-for-visibility-setting-label-for-trusted)

PR #1987, opened 2023-06-14
fix: Replace filled with outlined icons in professional activities

DanielPalafox:


DanielPalafox:
[[QA] Use material outlined icons for all professional activities](https://trello.com/c/f2XJ01YG/8668-qa-use-material-outlined-icons-for-all-professional-activities)

src/app/cdk/panel/panels/panels.component.ts
```diff
@@ -181,6 +185,9 @@ export class PanelsComponent implements OnInit {
   }
 
   ngOnInit(): void {
+    this._togglz
```
leomendoza123, line 188: Can we do this subscription only of `type='professionalActivitiesTogglz'` so we avoid so many subscribers? 

Also it might be a good idea to use `pipe(take(1))` when we use `getStateOf` since subscription is never completed by itself.

src/app/record/components/affiliation-stack/affiliation-stack.component.ts
```diff
@@ -190,5 +194,9 @@ export class AffiliationStackComponent implements OnInit {
     return false
   }
 
-  ngOnInit(): void {}
+  ngOnInit(): void {
+    this._togglz
+      .getStateOf('PROFESSIONAL_ACTIVITIES')
+      .subscribe((value) => (this.professionalActivitiesTogglz = value))
```
leomendoza123, line 200: The same comment on the other file applies here as well.

leomendoza123 (changes requested):
2 small changes to reduce the amount of not required subscriptions

src/app/cdk/panel/panels/panels.component.ts
DanielPalafox: makes sense I will do it

DanielPalafox:
> 2 small changes to reduce the amount of not required subscriptions

Done! @leomendoza123

amontenegro (approved):

PR #1989, opened 2023-06-19
Add ready for indexing check

leomendoza123:


leomendoza123:
[no-index is not being set for records without an OAuth token](https://trello.com/c/5z6Nv0KH/51-no-index-is-not-being-set-for-records-without-an-oauth-token)

amontenegro (approved):

PR #2009, opened 2023-07-09
fix: Override innerText set by OneTrust external script in link

DanielPalafox:


DanielPalafox:
[Access the Onetrust widget through a link in the site footer](https://trello.com/c/kr6xDzkD/8659-access-the-onetrust-widget-through-a-link-in-the-site-footer)

src/app/layout/footer/footer.component.ts
```diff
@@ -23,7 +23,11 @@ export class FooterComponent implements OnInit {
     })
   }
 
-  ngOnInit() {}
+  ngOnInit() {
+    setTimeout(() => {
+      document.getElementById('ot-sdk-btn').innerText = $localize`:@@footer.cookieSettings:Cookie Settings`
```
leomendoza123, line 28: Dani let's use the window provider to get the document object

leomendoza123: Also might be a good idea to check if the element exists, before trying to update the `innerText`

DanielPalafox: Done! @leomendoza123

leomendoza123 (approved):

PR #2012, opened 2023-07-11
Fix robots for ready for index users

leomendoza123:


leomendoza123:
[[PROD] No index tag is set to all records](https://trello.com/c/5cmaFcMs/8723-prod-no-index-tag-is-set-to-all-records)

amontenegro (approved):

PR #2038, opened 2023-08-10
Take the providerId from signinData if the entityDisplayName not available

Camelia-Orcid:


leomendoza123 (approved):

leomendoza123 (changes requested):
I am not super familiar with this session JSON. But is a bit confusing to me why we need to check if `this.entityDisplayName` is not falsy before re-assigning to a different value. 

Seems like your code would be equivalent to:

```
if (session.thirdPartyAuthData.entityDisplayName) { 
          this.entityDisplayName = this.signInData.providerId;
}
```

Camelia-Orcid:
We only want to set the this.entityDisplayName  to the providerId only if this.entityDisplayName = session.thirdPartyAuthData.entityDisplayName is nul or empty . Can you please review it.

leomendoza123 (approved):

PR #2064, opened 2023-09-28
feature: Add new source sorting option for affiliations, funding and …

DanielPalafox:
…works


src/environments/environment.local.ts
```diff
@@ -8,7 +8,7 @@ export const environment = {
   BASE_URL: '',
   INFO_SITE: 'https://info.orcid.org/',
   GOOGLE_ANALYTICS_TESTING_MODE: true,
-  GOOGLE_TAG_MANAGER: 'GTM-0000000',
+  GOOGLE_TAG_MANAGER: 'GTM-M9Z7DHS',
```
leomendoza123, line 11: we need to revert this right?

leomendoza123 (approved):

src/environments/environment.local.ts
DanielPalafox: Indeed

PR #2069, opened 2023-10-04
feature: Add `SOURCE_SORTING` togglz

DanielPalafox:


src/app/record/components/funding-stacks-groups/funding-stacks-groups.component.ts
```diff
@@ -40,16 +41,26 @@ export class FundingStacksGroupsComponent implements OnInit {
...
+    private _togglz: TogglzService
   ) {}
 
   ngOnInit(): void {
+    this._togglz
+       .getStateOf('SOURCE_SORTING')
+       .pipe(take(1))
+       .subscribe((value) => (this.sourceSorting = value))
     this.getRecord()
+    if (this.sourceSorting) {
```
leomendoza123, line 61: Since this `if (this.sourceSorting)` is out the subscription scope. It might have a unstable behaviour. Since the subscription might or might not have assigned the final value for `sourceSorting`. 

I think we can just move the if inside the subscription, so we now for sure what is the flag status.

src/app/record/components/work-stack-group/work-stack-group.component.ts
```diff
@@ -163,6 +171,9 @@ export class WorkStackGroupComponent implements OnInit {
     this._platform.get().subscribe((platform) => {
       this.platform = platform
     })
+    if (this.sourceSorting) {
```
leomendoza123, line 174: same here

leomendoza123 (changes requested):

src/app/record/components/funding-stacks-groups/funding-stacks-groups.component.ts
DanielPalafox: Great catch! You are right this togglz should be handled differently than the others. I will make the changes. Thanks

leomendoza123 (approved):

PR #2078, opened 2023-10-17
Revert "Angular update 15 16"

leomendoza123:
Reverts ORCID/orcid-angular#2077


leomendoza123:
Reverting as it was decided by product and qa, we will wait until next sprint to test this.

PR #2101, opened 2023-12-04
fix: Update additional email message error

DanielPalafox:


DanielPalafox:
[[QA] Apply dynamic email validation to additional email field in new register form](https://trello.com/c/EXyIW7p0/8951-qa-apply-dynamic-email-validation-to-additional-email-field-in-new-register-form)

src/app/register2/components/form-personal-additional-emails/form-personal-additional-emails.component.html
```diff
@@ -29,12 +29,12 @@
     <mat-error
       *ngIf="
         (additionalEmailsTouched &&
```
leomendoza123, line 31: MatError will be show only if the control was touched AND it has an error. 

With the parentheses changes you are showing the error immediately if the pattern is not correct, right? I believe we want to wait until the user focuses out the control to show the error.

leomendoza123 (changes requested):

amontenegro (approved):

PR #2115, opened 2023-12-08
8955-qa-registration-style-layout-and-content-issues

leomendoza123:


DanielPalafox:
[[QA] 'Existing email' notice panel and inline error](https://trello.com/c/XWBDcftY/8957-qa-existing-email-notice-panel-and-inline-error)

PR #2116, opened 2023-12-08
fix: Only display `email already associated` inline error if the user…

DanielPalafox:
… clicks on `next` button


DanielPalafox:
https://trello.com/c/XWBDcftY/8957-qa-existing-email-notice-panel-and-inline-error

leomendoza123 (approved):

PR #2135, opened 2024-01-15
8934 registration add an affiliation with typeahead

leomendoza123:


src/app/core/register2/register2.form-adapter.ts
```diff
@@ -116,20 +116,55 @@ export function Register2FormAdapterMixin<T extends Constructor<any>>(base: T) {
...
     }
 
+    formGroupToAffiliationRegisterForm(formGroup: UntypedFormGroup) {
+      const value = formGroup.controls['organization'].value
+      const departmentName = formGroup.controls['departmentName'].value
+      const roleTitle = formGroup.controls['roleTitle'].value
+      const startDateGroup = formGroup.controls['startDateGroup'].value
+
+      if (typeof value === 'string') {
+        return { affiliationName: { value } }
```
DanielPalafox, line 126: @leomendoza123 Affiliation type is missing here

DanielPalafox (changes requested):

src/app/core/register2/register2.form-adapter.ts
leomendoza123: Here is the card https://trello.com/c/kbiBtr8q/8934-registration-add-an-affiliation-with-typeahead

leomendoza123: Yes so I always assign the `employment` value to the type. It should always be `employment`

DanielPalafox (approved):

PR #2141, opened 2024-01-24
fix: Remove global navigation from registry

DanielPalafox:


DanielPalafox:
[Remove global navigation from the registry](https://trello.com/c/6rbD6ut6/9037-remove-global-navigation-from-the-registry)

leomendoza123 (approved):

PR #2147, opened 2024-02-07
fix: Add missing full stop in text and add missing error codes

DanielPalafox:


DanielPalafox:
[Registration - Unclaimed, deactivated and existing record notice panels](https://trello.com/c/5wDpHvGd/9035-registration-unclaimed-deactivated-and-existing-record-notice-panels)

PR #2149, opened 2024-02-08
fix: Add inline error message to additional email form

DanielPalafox:


DanielPalafox:
[[QA] Make it possible to display inline error for both primary and additional email fields in the registration form](https://trello.com/c/OjYLxei9/9047-qa-make-it-possible-to-display-inline-error-for-both-primary-and-additional-email-fields-in-the-registration-form)

amontenegro (approved):

PR #2175, opened 2024-02-29
fix: Redirec to signout instead of info site website

DanielPalafox:


DanielPalafox:
[[PROD] Sign out link in homepage header banner doesn't work](https://trello.com/c/9pzP6dt7/9102-prod-sign-out-link-in-homepage-header-banner-doesnt-work)

PR #2177, opened 2024-03-04
style: Fix record misalignment padding

DanielPalafox:


DanielPalafox:
[[QA] Record header misalignment](https://trello.com/c/jy3n28xO/9104-qa-record-header-misalignment)

PR #2227, opened 2024-04-08
Transifex

leomendoza123:


leomendoza123:
[Enable translations for the summary view in the public page](https://trello.com/c/CzNkKF7L/9191-enable-translations-for-the-summary-view-in-the-public-page)

PR #2240, opened 2024-04-23
fix: Hide record summary button if there is no public information ava…

DanielPalafox:
…ilable


DanielPalafox:
[Don't show record summary for records that are deactivated, locked, or deprecated](https://trello.com/c/u2GYhVAc/9177-dont-show-record-summary-for-records-that-are-deactivated-locked-or-deprecated)

amontenegro (approved):

PR #2244, opened 2024-04-24
fix: Read only organization field when editing affiliation if selected from db

DanielPalafox:


DanielPalafox:
[[QA] Organization fields are not read-only when editing affiliation or funding containing a disambiguated org](https://trello.com/c/4isr7CLk/9218-qa-organization-fields-are-not-read-only-when-editing-affiliation-or-funding-containing-a-disambiguated-org)

amontenegro (approved):

PR #2249, opened 2024-04-30
style: Update style to break word instead of break all text

DanielPalafox:


DanielPalafox:
[[QA] Stop words from breaking onto separate lines in the public page for mobile devices](https://trello.com/c/lMFeRkAX/9230-qa-stop-words-from-breaking-onto-separate-lines-in-the-public-page-for-mobile-devices)

PR #2254, opened 2024-05-02
fix: Remove organization data while clearing form

DanielPalafox:


DanielPalafox:
[[QA] Org info not cleared after removing org from typeahead](https://trello.com/c/upvET7uh/9233-qa-org-info-not-cleared-after-removing-org-from-typeahead)

amontenegro (approved):

PR #2257, opened 2024-05-09
Do not include query parameters in the ORCID id

amontenegro:


amontenegro:
[[PROD] Record summary in public page doesn't load when opened via link with tracking codes](https://trello.com/c/5y72Fggb/9242-prod-record-summary-in-public-page-doesnt-load-when-opened-via-link-with-tracking-codes)

leomendoza123 (approved):

PR #2271, opened 2024-05-29
fix: Remove sentenceCase in orcid logo text property

DanielPalafox:


DanielPalafox:
[Use the correct casing for the ORCID strap line in German](https://trello.com/c/BVsBuVwQ/9195-use-the-correct-casing-for-the-orcid-strap-line-in-german)

leomendoza123 (approved):

PR #2280, opened 2024-06-06
reduce header padding

auumgn:


auumgn:
[Adjust header padding to align with main page content](https://trello.com/c/KYBxzxvA/9267-adjust-header-padding-to-align-with-main-page-content)

PR #2308, opened 2024-07-24
improve dropdown activation logic

auumgn:


auumgn:
[[QA] Notification recipient address styling is different when there's more than one unverified address](https://trello.com/c/PE2jpljv/9322-qa-notification-recipient-address-styling-is-different-when-theres-more-than-one-unverified-address)

amontenegro (approved):

PR #2313, opened 2024-07-24
create extended dates and add timestamps to emails

auumgn:
I only used the extended date on the emails because the timestamps are only being returned in the emails endpoint


amontenegro (approved):

auumgn:
[[QA] Display emails addresses in descending order of created date](https://trello.com/c/rsCl20QO/9320-qa-display-emails-addresses-in-descending-order-of-created-date)

PR #2325, opened 2024-08-13
If the email address is not verified, we should not consider its visi…

amontenegro:
…bility for the domain visibility


amontenegro:
[[QA] Visibility setting of unverified emails is factored in when working out baseline visibility for associated domain](https://trello.com/c/8zKq9z0T/9348-qa-visibility-setting-of-unverified-emails-is-factored-in-when-working-out-baseline-visibility-for-associated-domain)

Camelia-Orcid (approved):

PR #2327, opened 2024-08-16
Fix bug on the logic to disable domain visibility

amontenegro:


amontenegro:
[[QA] Visibility setting of unverified emails is factored in when working out baseline visibility for associated domain](https://trello.com/c/8zKq9z0T/9348-qa-visibility-setting-of-unverified-emails-is-factored-in-when-working-out-baseline-visibility-for-associated-domain)

PR #2350, opened 2024-09-12
display both domains and emails on the public page

auumgn:


auumgn:
[Display both email addresses and domains in the public record](https://trello.com/c/JaL9ZRsL/9389-display-both-email-addresses-and-domains-in-the-public-record)

amontenegro (approved):

PR #2374, opened 2024-10-16
display orcid as the source if name is not public

auumgn:


auumgn:
[[QA] Email source display for records with private name](https://trello.com/c/g2Cqids6/9446-qa-email-source-display-for-records-with-private-name)

amontenegro (approved):

PR #2384, opened 2024-10-30
display name or orcid id when source id is missing

auumgn:


auumgn:
[Some self asserted email addresses are not showing a source](https://trello.com/c/qiIq8UrO/9460-some-self-asserted-email-addresses-are-not-showing-a-source)

PR #2488, opened 2025-04-01
load interstitials after loading user session in oauth

auumgn:


leomendoza123:
https://trello.com/c/FANKpDah/9724-qa-regression-test-oauth-and-openid-tests-not-getting-error-messages

PR #2494, opened 2025-04-03
do not show interstitial until a refresh is made

auumgn:
e.g. when you make all domains private and you've never seen the interstitial, you will need to refresh to see it instead of it appearing immediately


leomendoza123 (approved):

auumgn:
[[QA] Interstitial is always shown on OAuth if the local storage flag is not set](https://trello.com/c/tqZmdQJl/9729-qa-interstitial-is-always-shown-on-oauth-if-the-local-storage-flag-is-not-set)

PR #2495, opened 2025-04-03
fix get and set interstitial flow

auumgn:


auumgn:
[[QA] Interstitial is always shown on OAuth if the local storage flag is not set](https://trello.com/c/tqZmdQJl/9729-qa-interstitial-is-always-shown-on-oauth-if-the-local-storage-flag-is-not-set)

amontenegro (approved):

PR #2502, opened 2025-04-16
fix affiliation interstitial selector

auumgn:


auumgn:
[[QA] Affiliations interstitial shown during OAuth](https://trello.com/c/jzT2YvAd/9752-qa-affiliations-interstitial-shown-during-oauth)

PR #2507, opened 2025-04-23
Form params should be html encoded

amontenegro:


amontenegro:
[[Auth server][Phase 1] HTML encode form parameters on sign in](https://trello.com/c/pxfcXkGP/9758-auth-serverphase-1-html-encode-form-parameters-on-sign-in)

PR #2515, opened 2025-05-08
Send the user to the oauth authorize page after sign in when needed

amontenegro:


amontenegro:
[[QA] oauth sign in must send the request to the oauth link and not to my-orcid](https://trello.com/c/z8HR2yzD/9779-qa-oauth-sign-in-must-send-the-request-to-the-oauth-link-and-not-to-my-orcid)

PR #2523, opened 2025-05-19
9803-truncate-extremely-long-user-names-in-the-global-header

leomendoza123:


leomendoza123:
https://trello.com/c/UIcUWwFB/9803-truncate-extremely-long-user-names-in-the-global-header

amontenegro (approved):

PR #2561, opened 2025-07-17
Add featured works section

auumgn:


amontenegro (approved):

src/app/cdk/panel/panel-data/panel-data.component.ts
```diff
@@ -16,6 +16,10 @@ export class PanelDataComponent implements OnInit, OnDestroy {
 
   @HostBinding('class.orc-font-body-small') fontSmall = true
   @HostBinding('class.border-bottom') borderBottomClass = false
+  @HostBinding('class.featured') featuredClass = false
+  @Input() set featured(featured: boolean) {
```
leomendoza123, line 20: (Suggestion optional): Ideally this "panel data input" should have a more generic name so it can be reused for other features without causing confusion.

For example, we have the input "borderBottom", which is an explicit name that clearly indicates its effect on a generic component. It's not tied to a single feature.

src/app/cdk/panel/panel/panel.component.ts
```diff
@@ -87,6 +88,8 @@ export class PanelComponent implements OnInit {
   @Input() selectAll = false
   @Input() checkbox = false
   @Input() panelTitle: string
+  @Input() hasFeaturedIndex: boolean = false
```
leomendoza123, line 91: Do we really need both the `hasFeaturedIndex` input and the `featured-works` type input? Could we keep only one of those?

PR #2568, opened 2025-07-24
update featured works section on changes

auumgn:


src/app/core/record-works/record-works.service.ts
```diff
@@ -165,6 +165,12 @@ export class RecordWorksService {
    * @param options
    */
   getFeaturedWorks(options: UserRecordOptions): Observable<WorksEndpoint> {
+    options.pageSize = options?.pageSize || DEFAULT_PAGE_SIZE
```
leomendoza123, line 168: Seems like the `options` set here on the level of the class would overwrite the options set by the `getWorks`. 

Which I think could interfier if a user tries for instance to `updateVisibility` functionality

leomendoza123 (approved):

auumgn:
Part of https://app.clickup.com/t/9014437828/PD-1899

auumgn:
Bibtex change part of https://app.clickup.com/t/9014437828/PD-1968

PR #2569, opened 2025-07-28
${PD-1937} Handle the redirect uri from server vs the redirect uri stored in loc…

amontenegro:
…al memory


amontenegro:
Task linked: [PD-1937 [Auth server] Registration and reactivation fails when the authorization togglz is ON](https://app.clickup.com/t/9014437828/PD-1937)

PR #2576, opened 2025-07-31
${PD-1920} refactor feature works section

auumgn:
there is now a separate endpoint for fetching featured works which is not relying on pages and groups -- an array of works is returned instead

as a result a new component had to be introduced instead of using the existing logic for displaying works


auumgn:
also covers https://app.clickup.com/t/9014437828/PD-1984

PR #2627, opened 2025-09-15
${PD-2377} disable manage featured works button if there are no works

auumgn:


amontenegro:
Task linked: [PD-2377 Manage featured works button should be disabled when there are no works](https://app.clickup.com/t/9014437828/PD-2377)

PR #2631, opened 2025-09-15
${PD-2382} fix persistent featured state when searching works

auumgn:


amontenegro:
Task linked: [PD-2382 Issue with featured works search](https://app.clickup.com/t/9014437828/PD-2382)

PR #2633, opened 2025-09-16
${PD-2155} make header size adjustable in work details

auumgn:


amontenegro:
Task linked: [PD-2155 Heading structure in featured work expanded item panel not according to spec](https://app.clickup.com/t/9014437828/PD-2155)

PR #2820, opened 2026-04-16
PD-5449

cryptalith:


amontenegro:
Task linked: [PD-5449 Investigate increase in http_error events after OAuth release](https://app.clickup.com/t/9014437828/PD-5449)

PR #2831, opened 2026-05-08
PD-5529

cryptalith:


amontenegro:
Task linked: [PD-5529 [PROD] Double call to the authorize endpoint on re-authorization](https://app.clickup.com/t/9014437828/PD-5529)

PR #2888, opened 2026-07-29
PD-5781 Fix peer review i18n in print view

cryptalith:
[PD-5781](https://orcid.clickup.com/t/9014437828/PD-5781)

## Problem

The peer review heading is the only print-view string with placeholders, and it was broken in **every** locale — not falling back to English, but rendering truncated with both counts gone:

| | rendered |
|---|---|
| before | `Revisión por pares ($` |
| after | `Revisión por pares (7 revisiones de 3 publicaciones/concesiones)` |

`@angular/localize` emitted **no diagnostic**, so this shipped silently.

## Root cause

`fetch-orcid.js` is a plain asset, so `ng extract-i18n` never sees it. `normalize-xlf.prebuild.ts` parses the `$localize` calls out of the file instead, and its placeholder converter only matched UPPERCASE names:

```js
rawSource.replace(/\$\{[^}]*\}:([A-Z0-9_]+):/g, '{$$$1}')
```

The source uses camelCase (`:reviewCount:`), so nothing matched and the raw `${reviewsCount}:reviewCount:` text went verbatim into the XLF.

Fixing the name pattern alone would **not** have worked. `Xliff1TranslationParser` only recognises real `<x id="..."/>` elements, so the `{$NAME}` text form that converter aimed for parses as literal content, leaves the message with zero placeholders, and truncates the translation at the first `$`.

## Changes

- Emit real XLIFF `<x id="..."/>` placeholder elements. These are carried through the xml2js `Builder` as sentinels, since it escapes any markup found in a text node; the sentinels never reach disk.
- Always regenerate `printView.*` units from `fetch-orcid.js` instead of only adding them when absent, so re-parsing a unit that already holds `<x/>` cannot degrade it. Verified idempotent across repeated runs.
- Convert the locale targets in place, preserving the placeholder **reordering** translators applied in `tr` and `zh-TW`.
- Add a build-time check: fails on leftover legacy syntax, warns when a target drops a placeholder (which inlines cleanly but silently drops the value).

`<x/>` is the standard XLIFF 1.2 form Angular already emits elsewhere, and Transifex already round-trips it here — 7 translated targets per locale contain it today. The previous raw-text form was the outlier.

## Verification

- All 16 translated locales render both counts, run through the real `makeEs2015TranslatePlugin` pass; `uk`/`ca` correctly fall back to English.
- Full suite: **721 specs, 0 failures**. I confirmed the peer review spec actually executes by temporarily breaking it.
- `messages.xlf` is a one-line change — the usual regeneration churn was kept out.

## Follow-ups (not in this PR)

- **Push to Transifex before merging** so translations stay on the `<x/>` form; the new check fails the build if a pull brings the old syntax back.
- Pre-existing: `writeXlf`'s xml2js round-trip does not preserve mixed content, so the ~32 Angular-extracted units in `messages.xlf` have their `<x/>` elements moved to the end of each source on every build. Confirmed independent of this change (the unmodified script from `main` reproduces it). Tracked separately.


amontenegro:
Task linked: [PD-5781 Fix Peer Review i18n in Print View](https://app.clickup.com/t/9014437828/PD-5781)

amontenegro (approved):

PR #2910, opened 2026-08-22
PD-6181 Note other verified emails on the password reset confirmation

cryptalith:
Updates the password recovery confirmation copy for [PD-6181](https://orcid.clickup.com/t/9014437828/PD-6181), now that the backend sends the reset link to every verified address on the account.

`We have sent a recovery email to <email>` → `We have sent a recovery email to <email> and any other verified email addresses on your account`

## What changed

- **A new i18n unit, `@@recovery.andAnyOtherVerifiedEmails`, appended as a sibling fragment.** `@@recovery.weHaveSent` is left byte-identical so its translations in all ~19 locales survive; the new fragment simply arrives untranslated and fills in on the next Transifex pull. Editing the existing source in place would have put every locale at risk.
- **The sentence is gated on `recoveryType === 'password'`** via a new `isPasswordReset` getter. The confirmation block is shared with the "remind me my ORCID iD" flow, which posts to `/forgot-id.json` and still emails a single address — showing the fan-out wording there would be a lie. The getter is safe to read at confirmation time: `recoveryType` is initialised to `'password'` and its radio group lives inside the `@if (!submitted)` branch, so it cannot change once submitted.
- Whitespace follows the existing `>&nbsp;<b>` idiom. The component does not set `preserveWhitespaces`, so a leading space inside the span would have been trimmed out of the extracted source.

## Translation artifacts

`yarn build:i18n` touched only `messages.xlf` and the three auto-stamped pseudo-locales (`xx`, `lr`, `rl`) — no real locale file is in the diff, which is the check that the deprecated properties path did not run.

## Testing

`ng test ng-orcid`: **757 passing**, no failures.

Three new specs on a component that previously had only `should create`: the sentence renders on the password path, is **absent** on the forgot-iD path, and the submitted email still renders in both modes. The positive spec was confirmed to fail without the template change.

Also checked in a browser against a dev server, by driving the component directly (`window.ng.getComponent(el)` to set `recoveryType` and flip `submitted`) rather than submitting the form — there was no backend behind the local proxy, so this exercises the rendered template and the `isPasswordReset` gate, not the real submit path. It confirms the markup keeps a non-breaking space on both sides of the bold email, and that the iD-recovery path shows the original sentence unchanged:

> We have sent a recovery email to **josiah.carberry@brown.edu** and any other verified email addresses on your account

## Deploy order

Ships after ORCID/ORCID-Source#7651. With the backend not yet live the page would promise a fan-out that is not happening; the reverse order is harmless.


amontenegro:
Task linked: [PD-6181 Send password reset emails to all verified email addresses](https://app.clickup.com/t/9014437828/PD-6181)

amontenegro (approved):

PR #2917, opened 2026-09-01
PD-6175 allow RRID as an identifier type for funded-by relationships

cryptalith:
The funded-by allow-list only accepted grant_number, proposal-id, uri, and doi. Adds rrid to the validator and updates the error text and its i18n source strings to mention RRID.


amontenegro:
Task linked: [PD-6175 Update 'Funded by' relationship type to accept RRID as an identifier type in UI and API](https://app.clickup.com/t/9014437828/PD-6175)

PR #2923, opened 2026-09-07
PD-13324 speed up CI

cryptalith:
[PD-13324](https://app.clickup.com/t/86bbw0bw0)


cryptalith:
https://orcid.clickup.com/t/9014437828/PD-13324

PR #2924, opened 2026-09-07
PD-13322 Guard against notifications with a null source

cryptalith:
Fixes the inbox crash blocking PD-5720.


cryptalith:
https://orcid.clickup.com/t/9014437828/PD-13322

amontenegro (approved):

PR #2927, opened 2026-09-07
PD-5388 restore the record summary link on records with no public information

cryptalith:
https://orcid.clickup.com/t/9014437828/PD-5388


cryptalith:
https://orcid.clickup.com/t/9014437828/PD-5388

amontenegro (approved):

